PR #681, opened 2013-08-26
IsotopeShipping->getSurcharge() now returns surcharge even if price is 0

der-On:
At least in germany the shipping cost must be displayed even if they are 0.


aschempp:
Implemented in 9aaf3ef011eefb4bbe9d4cd19feff07687296085

PR #1250, opened 2014-09-12
Allow modification of gallery template

dmolineus:
At the moment the gallery template can not be changed. It would be helpful to change the gallery template, e.g. for render the main picture another way then the gallery.


aschempp:
merged in https://github.com/isotope/core/commit/9a4788c1da6878de3bcdf05909bcdd8c76e442df

PR #1302, opened 2014-11-06
[WIP] Start refactoring the rules component

tristanlins:
I have change the rule discount into an `inputUnit` field and provide a new discount type `unit`.
With `unit` you can define somethink like `-0.5 units`, that means the discount is `0.5` of the item unit price. My problem is, that I have a scaled price, so a fixed discount amount make no sense.

Currently the most is just reformatting. No BC breaks.


aschempp:
Is this still WIP ?

tristanlins:
Depending on what you expect ;-D
I cannot continue this work in any way... sry.

aschempp:
I'm closing this for now as it's still WIP and won't get any progress. I'll have a look at it when we try to rewrite rules :)

PR #1311, opened 2014-11-17
Add download file hook.

tristanlins:
Allow developers to hook into the download file process.
In my case, I need to stamp PDFs with licensee informations before download.

The hook synopsis:
```php
public function myDownloadFile(
    \FilesModel $objFileModel,
    \Isotope\Model\Download $objDownload,
    \Isotope\Model\ProductCollectionDownload $objCollectionDownload
);
```


aschempp:
Be aware that the hook name, position and arguments have changed.

PR #1343, opened 2014-12-18
Do not show "added to cart" on ajax request.

tristanlins:
When sending the cart form with ajax, the message will not shown on the next page reload anymore.


aschempp:
What did you do to the redirect request? That would not make sense either for AJAX?

I feel like it would be way simpler to just add a new button with it's own callback that does the AJAX functionality?

PR #1355, opened 2015-01-27
Multilingual product alias

ghost:
Made the product alias multilingual

### DIFF
--- a/system/modules/isotope/dca/tl_iso_product.php
+++ b/system/modules/isotope/dca/tl_iso_product.php
@@ -344,7 +344,7 @@ $GLOBALS['TL_DCA']['tl_iso_product'] = array
             'sorting'               => true,
             'inputType'             => 'text',
             'eval'                  => array('rgxp'=>'alnum', 'doNotCopy'=>true, 'spaceToUnderscore'=>true, 'maxlength'=>128, 'tl_class'=>'w50'),
-            'attributes'            => array('legend'=>'general_legend', 'fixed'=>true, 'inherit'=>true),
+            'attributes'            => array('legend'=>'general_legend', 'fixed'=>true, 'inherit'=>true, 'multilingual'=>true),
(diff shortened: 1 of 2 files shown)


aschempp:
Your subquery would result in an error if there are multiple products/translations with the same alias… Also, it would result in duplicate content because the same product can potentially be found through multiple URLs.

binron:
There was another bug when the subquery returned a empty result.
Changed it from "id = (...)" to "id IN (...)"

Duplicate content shouldn't be a problem as the alias field values must be unique:
bildschirmfoto 2015-01-30 um 14 58 22
 
And if for some reason two or more aliases would be equal there is already a limit:
```
         $arrOptions = array_merge(
             array(
                 'limit'     => 1,
                 'return'    => 'Model'
             ),
             $arrOptions
         );
```

aschempp:
What do I do if I want the same alias for all languages? Or the same for english and german but not for french?

binron:
That's a good point.
We could add a "multilingual alias" checkbox field to the products
and set the multilingual attribute of the alias field dynamically in a callback.
What do you think?

aschempp:
That would not solve the problem of same alias for two languages?

binron:
Yes that's true but it solves the lack of translated product URL's in a pragmatic way:
- If the feature isn't activated nothing will change at all
- No need to update data structure of existing Shops
- SEO friendly product URL's

micbis:
@aschempp Any update on this?

xApep:
Any news with this issue?

micbis:
So has this been fixed?

aschempp:
no, I deleted the master branch, thats why this got closed. Obviously it's very old too, so it can't be merged.
Multilingual product aliases would be a feature (not a fix) that I haven't found a way to implement yet. Maybe we'll look into it for Isotope v3 (which there are no real plans for yet).

Issue #1334, opened 2014-12-04
Cart session data is not validated


tristanlins:
- Put some products into the cart.
- Add a coupon code 
- Continue the purchase process until the last step
- Stop here, go back to the cart and remove all products
- Put another product into the cart (the product must excluded from delivery and the coupon is not compatible to the new products)
- Continue the purchase process
- The delivery and coupon are still active
![isotope](https://cloud.githubusercontent.com/assets/343404/5300646/2dc42b2c-7bd2-11e4-91e4-918afd58e019.png)

aschempp:
I can't reproduce this in 2.1, I assume @tristanlins was using 2.2? Because there it would make sense with the new darft order…

tristanlins:
No, it is a 2.1 from ER2.

aschempp:
Unfortunately, I can't reproduce the problem. Let me know if you're still working on that project or who could assist me with. I'll close the issue in the meantime.

kuvb:
+1
I just tried and could easily reproduce it with the demo installation (just updated to 2.1.10 Build 7)

![isotope-checkout-download-shipping](https://cloud.githubusercontent.com/assets/5165551/6082253/4b11090e-ae20-11e4-91b6-7124d4abff9b.jpg)

tristanlins:
My customer tell me last week, he can reproduce this with the official isotope demo! I will have a look today for this and maybe I can provide a fix later.

tristanlins:
Please see https://github.com/isotope/core/pull/1363

kuvb:
The label "invalid" might be a little confusing :)
I used the following demo files of Isotope: isotope_demo_2.1.7.zip. Happened before and after the update.
With tristanlins pull, it gets validated correctly on the review page, however not in the shopping cart.

You can reproduce the behaviour the following way:

- Add download product and shipping product to cart
- You need a shipping method that costs some money of course
- Proceed to review page
- Remove shipping product from cart
=> You will only have one download product in the cart and the price for shipping added as shown above in the screenshots.

kuvb:
@aschempp Were you able to reproduce this?
Is there a more "solid" fix that I can apply? The one of tristanlins only removes the shipping charges on the review page, not in the shopping cart.

tristanlins:
We did a quick modification on the `Cart.php:87` to solve the problem that an unnecessary shipping is displayed in the card (the other half of #1363):

```php
          // Remove shipping if not required
[...]

PR #1363, opened 2015-02-08
Clean shipping method, if its not required for checkout.

tristanlins:
Partial fix for #1334.


aschempp:
See my comment in https://github.com/isotope/core/issues/1334#issuecomment-81555467

PR #1377, opened 2015-02-23
Make the Payment::isAvailable() method result customisable

tristanlins:
I split the single restriction checks of `Payment::isAvailable()` into separate methods and give every method a hook to overwrite it's logic. I need this to fulfill a customers requirement.

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/Payment.php
+++ b/system/modules/isotope/library/Isotope/Model/Payment.php
@@ -109,7 +109,7 @@ abstract class Payment extends TypeAgent
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


aschempp:
I think that's totally over-engineered, why not just create a custom payment method?

tristanlins:
> I think that's totally over-engineered, why not just create a custom payment method?

Because I need to overwrite every payment method then!
The customer have some special rules when a payment method is available or not. And theses rules are depending on the payment method and some custom product attributes. 
I think it is over-engineering and not future-proof if I need to rewrite every payment method, that still exist only to apply some custom rules...

PR #1382, opened 2015-02-26
Update OrderConditions.php

MoniqueHahnefeld:
Workaround for checkbox or other Form-Elements, which return an array.
https://community.contao.org/de/showthread.php?55862-Simple-Token-form_*-bei-Merhfachauswahl

### DIFF
--- a/system/modules/isotope/library/Isotope/CheckoutStep/OrderConditions.php
+++ b/system/modules/isotope/library/Isotope/CheckoutStep/OrderConditions.php
@@ -138,7 +138,8 @@ abstract class OrderConditions extends CheckoutStep
 
         foreach (array_keys($this->objForm->getFormFields()) as $strField) {
             if (isset($_SESSION['FORM_DATA'][$strField])) {
-                $arrTokens['form_' . $strField] = $_SESSION['FORM_DATA'][$strField];
+                    $arrTokens['form_' . $strField] = is_array($_SESSION['FORM_DATA'][$strField])?implode(', ', $_SESSION['FORM_DATA'][$strField]):$_SESSION['FORM_DATA'][$strField];
+       
             }
         }
 


aschempp:
maybe we should implement something like the notification center?
https://github.com/terminal42/contao-notification_center/blob/master/classes/tl_form.php#L118-L136

MoniqueHahnefeld:
Hey,
For the isotope case the Notification won't be send from this class.
The Tokens will be handled from here.
https://github.com/terminal42/contao-notification_center/blob/98023c8d78ffed906498d9cb2f2c789c4d19dcf1/library/NotificationCenter/Util/String.php#L83-L92
I already tried and found no solution. It is difficult, because of the assoc array and the nested called convertToText. Maybe @Toflar knows a better solution? But the PR is a good fallback till we have a solution for this problem I suppose.
What do you think?

aschempp:
I didn't mean to use that method, but I was asking if we should implement the same. The functionality is different though, as every checkbox will result in a separate token instead of comma separated values.

MoniqueHahnefeld:
Yes I understood, but I found no way how you can implement it in the notification center.
And I am not sure what do you mean, how to use the token, in which syntax. So the question is, is ist possible to implement and how it could be used in BE. I suppose like ##group_field_val## and if you have multiple checkbox than "##group_field_val0##, ##group_field_val1##, ##group_field_val2## ..." ??? So what do you exactly mean?
Then if the User has only checke val0 and val2, what's with the divider ', '?

I also think that in the conditions case into isotope there are no really other needed formats than the dividing with comma. But maybe we can use a lang-global for variation caused on language. (Don't know how the divider looks in Russian and so on)

aschempp:
This should be implemented in the notification center in general. Meanwhile I have added a hook in Isotope to customize the tokens (see https://github.com/isotope/core/blob/2.1.13/system/modules/isotope/library/Isotope/CheckoutStep/OrderConditions.php#L128)

PR #1385, opened 2015-03-01
Replacing explicit module check with hook

patric-eberle:
I replaced an out of date support for the i18nl10n module with a new hook.

### DIFF
--- a/system/modules/isotope/drivers/DC_ProductData.php
+++ b/system/modules/isotope/drivers/DC_ProductData.php
@@ -431,15 +431,16 @@ class DC_ProductData extends \DC_Table
         $this->objActiveRecord = $objRow;
 
         // Load and/or change language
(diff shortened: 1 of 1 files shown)


Toflar:
It's back end...shouldn't it be a callback rather than a hook then? Well, Contao itself is not consistent in that matter anyway... :)

aschempp:
I guess that's a good point, because callbacks are tied to a specific DCA where hooks are not…

patric-eberle:
True. I'm looking forward for your callback solution then and will try to create a workaround for my module in the meantime.

aschempp:
Are you the developer of i18nl10n? Or how do you know it's outdated?

patric-eberle:
Yes I'm the developer. The config value you're checking is no longer available since two versions.

aschempp:
Implemented in b2b658a3dc1f89e09e53c581e202e54bd32181a5

You should set `DCA => config => languages` in an onload_callback to customize the options.

PR #1396, opened 2015-03-16
Respect shipping exempt when calculating weight

tristanlins:
Otherwise the weight exclusion of shipping modules will not be correct for products that are exempted from shipping.

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/ProductCollection.php
+++ b/system/modules/isotope/library/Isotope/Model/ProductCollection.php
@@ -1222,7 +1222,7 @@ abstract class ProductCollection extends TypeAgent
      *
      * @return Scale
      */
-    public function addToScale(Scale $objScale = null)
+    public function addToScale(Scale $objScale = null, $blnForShipping = false)
     {
         if (null === $objScale) {
             $objScale = new Scale();
@@ -1234,6 +1234,10 @@ abstract class ProductCollection extends TypeAgent
             }
 
             $objProduct = $objItem->getProduct();
+
+            if ($blnForShipping && $objProduct->shipping_exempt) {
+                continue;
+            }
 
             if ($objProduct instanceof WeightAggregate) {
                 $objWeight = $objProduct->getWeight();
--- a/system/modules/isotope/library/Isotope/Model/Shipping.php
+++ b/system/modules/isotope/library/Isotope/Model/Shipping.php
@@ -100,7 +100,7 @@ abstract class Shipping extends TypeAgent
             return false;
         }
 
-        $objScale = Isotope::getCart()->addToScale();
+        $objScale = Isotope::getCart()->addToScale(null, true);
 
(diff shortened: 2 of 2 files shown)


tristanlins:
I'm sure there are more places where this check is missing...

aschempp:
Why would you add a weight to the product if it's exempted from shipping? I mean just don't add the value to the product / the field to the product type?

tristanlins:
The product is a dual product, with 2 variants. A print and a download version. But the weight is not set on the print variant, it is set on the main product.

aschempp:
> The product is a dual product, with 2 variants. A print and a download version. But the weight is not set on the print variant, it is set on the main product.

I think that's an incorrect configuration of your shop. Still I see your point, I'm just not sure if it makes sense to check for this. With the correct configuration the admin would be able to choose if a shipping-exempted product should still be calculated (for whatever reason), which is not possible with your change…

/cc @Toflar @qzminski any opinions?

Toflar:
Well, in this case I'm probably with @tristanlins here. It's the same product, just in an other variant. It has the same description, probably the same price, same author, same ISBN etc. One variant is just downloadable and the other is not. However, I think the implementation is wrong.

aschempp:
It's the same product, but a different type. If you would sell two different versions of a book, they would also have two different weights, so a config in the variant makes sense.

As an example: https://de.wikipedia.org/wiki/Taschenbuch
> Ein Taschenbuch […] ist ein kleinformatiges Buch mit gewöhnlich weichem Einband. Es wird […] meist einige Zeit nach den höherwertigen fest (engl. „Hardcover“) oder weich („Broschur“) gebundenen Ausgaben veröffentlicht.

Toflar:
Ah I misunderstood then.

> But the weight is not set on the print variant, it is set on the main product.

That's obviously a misconfiguration. The weight has to be set in the variant.

tristanlins:
What about collection by the customer (Selbstabholer)? 
An item that is excluded from shipping must not only downloadable.

aschempp:
That would be a shipping method not a product attribute?

tristanlins:
I disagree, in the past I had build a shop with another software. My customer sells wooden products. Some are available for shipping and some only for collection by the customer. But when you buy two products, one for shipping and one for collection by the customer you can still select another shipping method for the first product. If you say this is a product attribute, why does the "except from shipping" attribute exist? It make no sense then.

aschempp:
> If you say "collection by the customer" is a shipping method, not a product attribute, why does the "except from shipping" attribute exist? It make no sense then.

Because if you have only products that are not shipped, the shop does not ask for a shipping method at all. Again, the solution to your problem is very simple: do not enter a weight for the products not to be calculated. What's the problem with that? Even the field label is *Shipping weight* …

tristanlins:
I agree that my setup is wrong.
But the behavior of your field is still wrong. It is valid that you have products with weight (e.g. for price calculation), but these products are excepted from shipping.

Issue #1472, opened 2015-06-30
Picture select menu


jenda77:
Hello, 

I would like to have the possibility to add the pictures as radio buttons - it means picture select menu. I think it belongs to EFG. Would you like to implement it? It would be great. Thank you. 

Jan
![bildschirmfoto 2015-06-30 um 16 35 47](https://cloud.githubusercontent.com/assets/3715050/8433461/35dd63a6-1f46-11e5-9524-aaa3a9c515d1.png)

Toflar:
Pull requests are welcome :)

PR #1480, opened 2015-07-06
Picture select menu

jenda77:
Hello,

I would like to have the possibility to add the pictures as radio buttons - it means picture select menu. I think it belongs to EFG. Would you like to implement it? It would be great. Thank you.

Jan

https://github.com/isotope/core/issues/1472


aschempp:
You got this wrong. A pull request is meant that you provide the implementation of this feature for us and everyone else to use. Someone has to do the work, and there are other priorities on our list for now. So if you want this feature, pay a developer to do it for you and contribute back to the project by sending us the necessary changes.

jenda77:
I see. Thank you for the clarify

PR #1535, opened 2015-09-18
Added opportunity to localize the backend email data fields

Defcon0:


aschempp:
I've added this for Isotope 2.4 but changed the label array to be more specific in 78968c61768a6bbceb76f1c05493eb5b6db61459

PR #1618, opened 2016-03-02
Product Group edit for non admin users

tsarma:
Please consider this change. Non admin users can not move or delete product group even though these functions are assigned to their users group.

screen shot 2016-03-02 at 14 17 45


aschempp:
Fixed in 2be1ad0689c31411af76aa431a5db82d480a4b76

Issue #1473, opened 2015-07-01
Hook for Address update


Zeromax:
It would be cool to have a hook in the address module when a address has changed.
With the old and new data like the `$GLOBALS['TL_HOOKS']['updatePersonalData']` from Contao.

That would offer for example to send a notfication for changed addresses.

Is it to late for 2.3, cause its already in the RC?

aschempp:
Yes that's too late for 2.3, but certainly a good idea.

aschempp:
Implemented in #1625

PR #1625, opened 2016-03-14
Added hook triggered when address data is updated

qzminski:
See #1473

### DIFF
--- a/system/modules/isotope/docs/CHANGELOG-2.4.md
+++ b/system/modules/isotope/docs/CHANGELOG-2.4.md
@@ -11,3 +11,4 @@ Version 2.4.0-beta1 (201?-??-??)
 - Added hook to modify fields in checkout process
 - Added option to edit products that were added to the cart
 - Added config setting to define an order details module for backend view (#1578)
+- Added hook triggered when address data is updated (#1473)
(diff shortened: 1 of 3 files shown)


aschempp:
I don't think the hook should be triggered in `CartAddress`. The `CartAddress` module does not store in the member's address book. Or did I get something wrong?

qzminski:
I was unsure about that. Does it not store it here?

https://github.com/isotope/core/blob/master/system/modules/isotope/library/Isotope/Module/CartAddress.php#L134

aschempp:
Yeah, it potentially alters your default billing address. But it could also be a temporary cart address, we don't know. If you look at the example in #1473 I don't think it would make sense to send a notification in the cart address module.

qzminski:
Reverted in 2045fb3.

aschempp:
Now looking at it in detail, I wonder why we need a hook if there is the `onsubmit_callback`? @Zeromax

Zeromax:
I can not remember why this was needed. I have a new Job and I am not able to reproduce this all in my head... Sorry about that.

Is both available in the onsubmit_callback? Than maybe I missed this callback or it has to less information...

aschempp:
Well there's not reason not to add it anyway :D

Zeromax:
@aschempp thanks

Issue #1542, opened 2015-09-26
Validate price attribute in product and variant


aschempp:
It's not supported to have a price in the product and the variant. We should add a validation as this is something a lot of people do wrong.

aschempp:
I think that validation should be on the product type. Just something like a save_callback on `variant_attributes`.

aschempp:
Implemented in #1626

PR #1626, opened 2016-03-14
Validate the price attribute in product type against duplication

qzminski:
See #1542.

Should we also check for the ```baseprice``` field?


Toflar:
I think this would be better if it was an `eval` flag so it can be activated in whatever attribute you don't want to have configured in both, product and variant. I even think the core does have more than just "price", no?

aschempp:
I agree with @Toflar, this should be a field flag, so we can easily extend it if we find new fields that must not be duplicates. I can't come up with a clever name though :-(

qzminski:
I have changed that in 1e35404.

PR #1627, opened 2016-03-14
Added the responsive images feature

qzminski:
See #1423.

### DIFF
--- a/system/modules/isotope/dca/tl_iso_gallery.php
+++ b/system/modules/isotope/dca/tl_iso_gallery.php
@@ -173,7 +173,7 @@ $GLOBALS['TL_DCA']['tl_iso_gallery'] = array
             'label'                 => &$GLOBALS['TL_LANG']['tl_iso_gallery']['main_size'],
             'exclude'               => true,
             'inputType'             => 'imageSize',
-            'options'               => $GLOBALS['TL_CROP'],
+            'options'               => \System::getImageSizes(),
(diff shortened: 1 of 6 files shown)


system/modules/isotope/dca/tl_iso_gallery.php
```diff
@@ -218,7 +218,7 @@
             'label'                 => &$GLOBALS['TL_LANG']['tl_iso_gallery']['zoom_size'],
             'exclude'               => true,
             'inputType'             => 'imageSize',
-            'options'               => $GLOBALS['TL_CROP'],
+            'options'               => \System::getImageSizes(),
```
Toflar: Please use the options_callback so this stuff is lazy loaded :) The data is loaded in the FE everytime a Model instance is created and thus relevant for performance :)

aschempp:
Why do we call both `Image::create` and `Picture::create`?

qzminski:
```Image::create``` for BC reasons for custom ```iso_gallery_``` templates. ```Picture::create``` for new gallery templates. I copied this code from contao/core's ```Controller::addImageToTemplate``` method.

aschempp:
You copied? :see_no_evil: 

Why not use the image URL from `Picture::create`?

qzminski:
Apparently there is a reason why ```Controller::addImageToTemplate``` handles it that way.

aschempp:
Removed picture support for inline and elevateZoom gallery in 32d9188b706a1e511507ec223f76a1029770c9a8

Issue #1561, opened 2015-10-30
FileTree sorting


tsarma:
Please make an attribute of type FileTree sortable

aschempp:
Implemented in #1629

PR #1629, opened 2016-03-16
Made the FileTree attribute sortable

qzminski:
See #1561.


aschempp:
do we not need an option to enable or disable the feature?

qzminski:
In my opinion not. However we could also provide the ```isGallery``` flag which should make the sorting for galleries much convenient.

aschempp:
Do we have settings for extensions etc. then? Or would you enable `isGallery` if the image types are only gallery? Be aware that it is not always desired to have sorting. Is there a setting for the sorting of the files?

qzminski:
I would add another checkbox that turns on ```isGallery``` flag if enabled.

> Be aware that it is not always desired to have sorting. Is there a setting for the sorting of the files?

There was no setting before and now there is none either. Previously the files were displayed in the frontend in the order they were added to the widget, now they are sorted manually (still only on request though because if you don't sort them manually they behave the same).

aschempp:
Can we add a sorting configuration like it is available in a gallery? Like sort by name, date, etc? And a manual sorting option, and only then add the order field?

qzminski:
That would be the most fair deal. Just wondering, if you select the manual sorting option, would you then have to update the database manually to create that ```orderField``` that stores manual order? Or can attributes already handle the automatic update of db on each save?

aschempp:
Good question. As far as I remember, the database will be update automatically if an attribute changes its configuration. The *magic* will not delete any fields, but it should create the necessary ones. Not sure if it only handles the `field_name` though, please test that.

qzminski:
I have added the ```sortBy``` support and the new ```isGallery``` option.

Regarding the order field, it is always created alongside with the ```fileTree``` attribute default field. So when I add a new attribute of that type, two fields are created: ```files``` and ```files_order```. That's because the ```custom``` value of ```sortBy``` is the default one.

If you change the ```sortBy``` value of such attribute then you can manually drop the order field in the database update tool.

aschempp:
What if we would not set «manual» as the default option? Wouldn’t that be better?

qzminski:
Which option would you set by default then? The other thing is that we don't know if the order field would be then created automatically upon save (preceded by ```sortBy``` change to ```custom```). Frankly I would stay with the current solution.

aschempp:
I would simply choose *File name (ascending)* and move the Custom Order to the last position. I'm pretty sure the field will be added because Isotope performs an automatic sync of DB (except for deleting). But you should obviously test that ;-)

qzminski:
I have changed the default sorting in a232efb. It seems to work as expected, i.e. the order field is created automatically when the ```custom``` option has been chosen and the record saved.

PR #1634, opened 2016-03-22
CurrencySpace with a non-breaking space

michelbalzer:
siehe #1628

### DIFF
--- a/system/modules/isotope/library/Isotope/Isotope.php
+++ b/system/modules/isotope/library/Isotope/Isotope.php
@@ -325,9 +325,9 @@ class Isotope extends \Controller
         $strPrice    = static::formatPrice($fltPrice, $blnApplyRoundingIncrement);
 
         if ($objConfig->currencySymbol && $GLOBALS['TL_LANG']['CUR_SYMBOL'][$strCurrency] != '') {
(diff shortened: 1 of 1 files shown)


aschempp:
Thanks!

Two things:
 1. Why not use `&nbsp;`?
 2. We must only output a HTML character if the output type is HTML (see the method argument).

michelbalzer:
To 1.) Think I saw somewhere else in Isotope that the &# declaration was used. Otherwise there was no specific reason. Both should work.

To 2.) uhm, yes, my bad. The non-html version is used in non-html mails for example, right? We could directly put the character in it.

Von unterwegs gesendet

michelbalzer:
As I researched: &#160; is faster ;)

Von unterwegs gesendet

aschempp:
> The non-html version is used in non-html mails for example, right? We could directly put the character in it.

Correct, that's for plain text mails for example. Can you update the PR?

michelbalzer:
Uhm, which one do I have to replace?

aschempp:
Personally, I would prefer both. `&nbsp;` is easier to read for most users/developers, that should always be more important that theoretical performance. And we also must not use HTML in non-HTML mode.

aschempp:
Merged in 4aa8fdc08619571cf1e8a39c935704033fff2ae3

PR #1640, opened 2016-04-08
Security Update

craidler-radconsulting:
Datatrans changed security settings so the sign needs to be sha256 with the Datatrans HMAC hex being converted to bin before using it to generate the sign itself.


aschempp:
If we change this, it will affect existing installations, right?

craidler-radconsulting:
Deffo! So please feel free to change the push destination appropriatly or close PR.

aschempp:
can the user choose which hash algorithm to use in the Datatrans backend? Then we should probably provide the same options in Contao (similar as we do for Postfinance).

craidler-radconsulting:
Added select field for choosing the algorithm and the option to convert from hex2bin. Defaulted to md5 and 0, which will leave existing installation behaviour as is. I checked this on my system but i'd be happy if you could double check. Also i only added a german translation for it.

Toflar:
This should be tagged 2.4 then.

aschempp:
Merged in 50a3f975717cd5ae91fa566635bd09dcde9faf67

PR #1650, opened 2016-05-24
Add Template variable for the count of products

trcwebdesign:
With this change you can get at the template the count of the products including filtered
<<?php echo $this->hl; ?>><?php echo $this->headline; echo " (" , $this->countProducts ,")"  ?></<?php echo $this->hl; ?>>

### DIFF
--- a/system/modules/isotope/library/Isotope/Module/ProductList.php
+++ b/system/modules/isotope/library/Isotope/Module/ProductList.php
@@ -119,6 +119,9 @@ class ProductList extends Module
      */
     protected function compile()
     {
+       $productCount = 0;
+       $queryBuilder = null;
+
         // return message if no filter is set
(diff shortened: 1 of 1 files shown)


aschempp:
Hm, wouldn't it be as simple as `<?= count($this->products) ?>` in the list template?

mrothtrc:
Nope .. 
count($this->products) = 5 (Pagination value and current displayed products)
countProducts = 52 or 251 if no filter active
The variable countProducts is set before any pagination and merging

aschempp:
I have added template variables for all pagination information in 0769f375d8d1e7fb7cf35ddabfef37846f5174ff

aschempp:
I have added template variables for all pagination information in 0769f375d8d1e7fb7cf35ddabfef37846f5174ff

PR #1682, opened 2016-08-01
Added the gulp tasks to minify CSS and JS

qzminski:


aschempp:
Thanks! Would you agree we can ignore stuff like browserify?

qzminski:
We do not require any module directly in the code yet, so we can ignore it.

PR #1684, opened 2016-08-09
Fix value of a single checkbox attribute in backend

### LINKED ISSUE #1658: Single checkbox attribute issue
If you create the attribute "checkbox menu" with a single option created via Options Manager, then you won't be able to check and save that field in the product settings. The reason is that the options manager creates the options array including the option database record ID and the option label:

```php
$GLOBALS['TL_DCA']['tl_iso_product']['fields']['my_field']['options'] = [
    123 => 'My option',
];
```

That is correct for the multiple checkboxes but the single one gets **always** value=1, see:

https://github.com/contao/core/blob/master/system/modules/core/widgets/CheckBox.php#L188

Thus when you attempt to save the record you will get ```Invalid input: 1``` error message because there is no option with key ```1```.

qzminski:
Fixes #1658

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/Attribute/CheckboxMenu.php
+++ b/system/modules/isotope/library/Isotope/Model/Attribute/CheckboxMenu.php
@@ -66,7 +66,7 @@ class CheckboxMenu extends AbstractAttributeWithOptions implements IsotopeAttrib
         $options = parent::getOptionsForWidget($objProduct);
 
         // Make sure that the option value for single checkbox in backend is always "1" (#1658)
-        if (TL_MODE === 'BE' && !$this->multiple) {
+        if (!$this->isVariantOption() && !$this->isCustomerDefined() && !$this->multiple) {
             $options[0]['value'] = 1;
         }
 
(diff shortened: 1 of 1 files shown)


qzminski:
Ah should be to the ```develop``` obviously. Sorry about that.

aschempp:
I don't think that's the right solution. You can still add multiple options to that select menu, which would not make sense at all?

qzminski:
Huh? What select menu? I don't get what you said.

aschempp:
Sorry, I mean checkboxes obviously. For a single checkbox, we do not need options, because the field label is the option, right? So we should hide tl_iso_attributes.options and optionsSource if multiple is not checked?


Freundliche Grüsse

Andreas Schempp
terminal42 gmbh

qzminski:
You can't do that I am afraid as in the frontend the ```1``` value is not hardcoded contrary to the backend.

aschempp:
So we only need to do what I said if the checkbox is backend only (not *variant option* and not *customer defined*)?

qzminski:
Yup, I think so.

aschempp:
Can you adjust the PR then?

aschempp:
Fixed in f48882dc365bbbafa6b0d39ae2f792f496bf06a1

I've added a new "options source" if the field is not customer editable (backend only).
bildschirmfoto 2016-10-03 um 11 31 34

PR #1685, opened 2016-08-09
Fix the elevate zoom multiple usage

### LINKED ISSUE #1674: [2.3.5] "Elevate Zoom gallery" funktioniert nur beim ersten Produkt
Wenn "Elevate Zoom gallery" in der Galerie ausgewählt wird und man diese Galerie unter "Produktetypen" / "Galerie auflisten" auswählt, funktioniert der Zoom nur auf dem ersten Produkt.

Vielleicht hängt es damit zusammen, das wenn man Produkte erstellt und da ein Bild hochlädt, dass alle Bilder die selbe ID erhalten. 

[Siehe auch: ](https://community.contao.org/de/showthread.php?63119-Probleme-mit-Produkte-Varianten-und-Zoom-Anzeige)

qzminski:
Fixes #1674

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/Gallery/ElevateZoom.php
+++ b/system/modules/isotope/library/Isotope/Model/Gallery/ElevateZoom.php
@@ -37,9 +37,6 @@ class ElevateZoom extends Inline
     {
         parent::addImageToTemplate($objTemplate, $strType, $arrFile, $blnWatermark);
 
-        // Make sure ID is unique so the gallery can be used on product list (#1674)
-        $objTemplate->uid .= $this->product_id;
-
         if ($blnWatermark) {
             if ('main' === $strType) {
                 $GLOBALS['TL_JAVASCRIPT'][] = 'system/modules/isotope/assets/plugins/elevatezoom/jquery.elevateZoom-3.0.8.min.js';
--- a/system/modules/isotope/library/Isotope/Model/Gallery/ElevateZoom.php
+++ b/system/modules/isotope/library/Isotope/Model/Gallery/ElevateZoom.php
@@ -37,6 +37,9 @@ class ElevateZoom extends Inline
     {
         parent::addImageToTemplate($objTemplate, $strType, $arrFile, $blnWatermark);
 
+        // Make sure ID is unique so the gallery can be used on product list (#1674)
+        $objTemplate->uid .= $this->product_id;
+
         if ($blnWatermark) {
             if ('main' === $strType) {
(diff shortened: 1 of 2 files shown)


aschempp:
Looks like that would also be an issue in any other gallery that uses the "uid"? Why is `uid` not unique by default?

qzminski:
The uid is not unique because it takes the object hash:

https://github.com/isotope/core/blob/master/system/modules/isotope/library/Isotope/Model/Gallery/Inline.php#L71

aschempp:
So that's wrong then. The `uid` must be unique, but it must be consistent for the same gallery and same product. Only fixing ElevateZoom would not fix this in InlineGallery then.

qzminski:
I have changed that in b93334a.

aschempp:
I'm sorry but that won't work either. First of all, module ID 5 and product ID 75 would result in the same as module ID 57 and product ID 5. Even worse, a second call to the same method would not get the same `uid`. 

How about something like this?

```php
$objTemplate->uid = md5('gallery' . $this->id . '::product' . $this->product_id);
```

We should also check if there is a product ID at all, I think that's not mandatory, so we might only use `md5('gallery'.$this->id)` in that case.

qzminski:
What if I have the product reader and the product list on the same page consisting the same product with the same gallery? That would also result in duplicate ID, no?

That's the reason I have used ```uniqid()``` method so upon each gallery generation the uid is truly unique. Remember ```$this->id``` refers to the gallery ID (model) and not to the module ID.

aschempp:
You're right. But a subsequent call to `addToTemplate` is also possible. Do we have the module ID to append, or maybe we can add that somehow?

qzminski:
The module ID is too precise here I think as the product could also be potentially generated by a custom content element. So I would rather introduce the ```setReferenceId()``` method in gallery that could be still used by frontend module but also by others.

Or generate that unique ID upon the gallery object construct? It's a model after all so it should be in the registry and created only once per runtime.

aschempp:
I think using the module ID or whatever source identifier we have would be correct. We already do have a config array that contains the module. Most of the time it would not matter anyway.

Regarding all other options, the won't work.
- Generating unique ID on object construction wont work, as it's a model in the registry and the same model is used for all galleries (very stupid design fault…).
- That's why the object hash does not work either, the issue we currently have.
- Using a generated value on each access does not work, because subsequent calls to generate the main image and the gallery must result in the same hash.

qzminski:
Then I would have something like ```$referenceId``` in the gallery class with a method that can be used to set it. What do you think?

aschempp:
I think I found a very simple solution 😂 

@qzminski can you please verify b487751964cc335f244bef5d5bb1c3d3f3c5bf72 and close this issue if solved?

qzminski:
Seems to work 👍

PR #1687, opened 2016-08-09
Clean the parameter value in the Payone payment method

### LINKED ISSUE #1256: Escape &nbsp; in Product Names for PAYONE
Good morning,

a new week, a new PAYONE 'challenge'…

I discovered that there is a problem with the current implementation for PAYONE that leads to incorrect hash calculation. We use product names with [nbsp] to prevent ugly line wraps (e.g. in front of units). These are rendered as the _&#x00A0_; character in HTML (as value of the hidden input fields). Browsers do not encode this value for some reason (!), whereas ```urlencode``` does correctly convert it to _%C2%A0_.
To circumvent this, I added a simple ```str_replace``` for [nbsp] to a regular whitespace.

Line 128 of Model/Payment/Payone.php:
```php
            $arrData['de[' . $i . ']']   = specialchars(str_replace('[nbsp]', ' ', $objItem->getName() . $strOptions));
```

This solves it for me and PAYONE now accepts the requests. I guess one could find a cleaner solution, but I already invested two hours to find the cause.

qzminski:
Fixes #1256

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/Payment/Payone.php
+++ b/system/modules/isotope/library/Isotope/Model/Payment/Payone.php
@@ -134,10 +134,7 @@ class Payone extends Postsale implements IsotopePayment
             $arrData['id[' . ++$i . ']'] = $objItem->getSku();
             $arrData['pr[' . $i . ']']   = round($objItem->getPrice(), 2) * 100;
             $arrData['no[' . $i . ']']   = $objItem->quantity;
-            $arrData['de[' . $i . ']']   = specialchars(
-                \StringUtil::restoreBasicEntities($objItem->getName() . $strConfig),
-                true
-            );
+            $arrData['de[' . $i . ']']   = $this->escapeParameterValue($objItem->getName().$strConfig);
         }
 
         foreach ($objOrder->getSurcharges() as $k => $objSurcharge) {
@@ -148,9 +145,8 @@ class Payone extends Postsale implements IsotopePayment
             $arrData['id[' . ++$i . ']'] = 'surcharge' . $k;
             $arrData['pr[' . $i . ']']   = $objSurcharge->total_price * 100;
             $arrData['no[' . $i . ']']   = '1';
-            $arrData['de[' . $i . ']']   = $objSurcharge->label;
+            $arrData['de[' . $i . ']']   = $this->escapeParameterValue($objSurcharge->label);
         }
-
 
         ksort($arrData);
         // Do not urlencode values because Payone does not properly decode POST values (whatever...)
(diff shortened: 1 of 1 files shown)


system/modules/isotope/library/Isotope/Model/Payment/Payone.php
```diff
@@ -168,4 +164,22 @@ function(&$option) {
...
+     *
+     * @return string
+     */
+    protected function escapeParameterValue($value)
+    {
+        $value = strip_insert_tags(strip_tags($value));
+        $value = \StringUtil::restoreBasicEntities($value);
+        $value = html_entity_decode($value);
+        $value = str_replace("\xC2\xA0", ' ', $value); // strip &nbsp;
+        $value = str_replace('"', '\'', $value); // replace double quotes with single quotes
```
aschempp: Why do we need single quotes?

qzminski: To prevent from breaking the HTML ```value``` attribute in a hidden field.

aschempp: Ah, that's what I use `StringUtil::specialchars` usually. One or the other would be ok, but why not use `\"` then?

qzminski: Does HTML support escaping characters by backslash? 😄

aschempp: Absolutely. And you're already doing that with `\'`? 😂

qzminski: Absolutely not you mean 😉  The ```\'``` is escape only in PHP so it results as a single quote in HTML. You can't escape the form value as follows:

```html
<input type="hidden" name="foobar" value="my \"super\" product name">
```

aschempp: So maybe `StringUtil::specialchars()` is the way to go?

qzminski: Can you send me the URL to that class&method on Github? I can't find it neither in Contao core nor in Haste.

aschempp: Maybe that's only available in Contao 4. Just use the `specialchars()` method in Contao.

qzminski: But then it will be converted to the HTML entity which may be a problem just like ```&nbsp;```.

aschempp:
Fixed in https://github.com/isotope/core/commit/8410fd423f6952f55e46ce30bed95ca996df8c1e

PR #1689, opened 2016-08-10
Fix displaying and validating subdivisions with subgroups

### LINKED ISSUE #1678: Subdivisions with subgroups not validating or displaying properly.
When using some subdivisions (GB is the best example) which have another dimension to their array (for sub-subdivisons), the validation on checkout does not let a user proceed. (Invalid Input).

I see that this may be an issue in the FormConditionalSelect menu, as there is a custom isValidOption on ConditionalSelect which seems to clear up that issue, but that still leaves the issue where in `Isotope\Model\Address` line 198 there is still not a check for another dimension in the array.

qzminski:
Fixes #1678

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/Address.php
+++ b/system/modules/isotope/library/Isotope/Model/Address.php
@@ -203,7 +203,21 @@ class Address extends \Model
 
                 list($country, $subdivion) = explode('-', $this->subdivision);
 
-                $arrTokens['subdivision']      = $arrSubdivisions[strtolower($country)][$this->subdivision];
+                foreach ($arrSubdivisions[strtolower($country)] as $v) {
+                    // Support multi-dimensional arrays (e.g. subdivisions of United Kingdom)
+                    if (is_array($v)) {
+                        foreach ($v as $vv) {
+                            if ($vv[$this->subdivision]) {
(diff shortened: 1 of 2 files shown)


system/modules/isotope/library/Isotope/Model/Address.php
```diff
@@ -203,7 +203,21 @@ public function getTokens($arrFields = null)
...
+                    // Support multi-dimensional arrays (e.g. subdivisions of United Kingdom)
+                    if (is_array($v)) {
+                        foreach ($v as $vv) {
+                            if ($vv[$this->subdivision]) {
+                                $arrTokens['subdivision'] = $vv[$this->subdivision];
+                                break;
+                            }
+                        }
+                    } else {
+                        $arrTokens['subdivision'] = $arrSubdivisions[strtolower($country)][$this->subdivision];
```
aschempp: This can't be right. We're `foreach`ing the array but then not use any of it's foreach-value in the loop? 

Did you test this? On the first spot it seems wrong to have two `foreach`?

qzminski: I have tested it and it works as expected. We are using the foreach-value in the loop, see ```$v``` on line 209. If that is not an array on the other hand, then the loop is immediately broke in 217.

PR #1690, opened 2016-08-12
Do not display product in breadcrumb if the page type is 404

### LINKED ISSUE #1673: [2.3.4] Breadcrumb bei Fehlerseite wenn Produkt auf nicht veröffentlichter Seite
Wenn ein Produkt aufgerufen wird, das sich auf einer nicht veröffentlichten Seite befindet, wird korrekt ein 404-Fehler erzeugt.

Jedoch wird das Produkt auf der Fehlerseite in der Breadcrumb-Navigation angezeigt, dies dürfte jedoch nicht sein, denn nicht gefundene Elemente werden in der Breadcrumb-Navigation grundsätzlich nicht angezeigt.

qzminski:
Fixes #1673

### DIFF
--- a/system/modules/isotope/library/Isotope/Frontend.php
+++ b/system/modules/isotope/library/Isotope/Frontend.php
@@ -633,8 +633,7 @@ class Frontend extends \Frontend
         /** @var \PageModel $objPage */
         global $objPage;
 
-        if ($objPage->type === 'error_404'
-            || $objPage->type === 'error_403'
+        if (!($objPage->type instanceof \PageRegular)
             || !($alias = Input::getAutoItem('product', false, true))
             || ($objProduct = Product::findAvailableByIdOrAlias($alias)) === null
         ) {
--- a/system/modules/isotope/library/Isotope/Frontend.php
+++ b/system/modules/isotope/library/Isotope/Frontend.php
@@ -637,27 +637,30 @@ class Frontend extends \Frontend
 
                 /** @var \PageModel $objPage */
                 global $objPage;
-                global $objIsotopeListPage;
 
-                $last = count($arrItems) - 1;
+                if ($objPage->type !== 'error_404') {
+                    global $objIsotopeListPage;
 
-                // If we have a reader page, rename the last item (the reader) to the product title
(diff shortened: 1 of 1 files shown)


aschempp:
Please rewrite to use early exists. 

```php
if (if ($objPage->type === 'error_404')) {
    return;
}
```

Also apply to the other existing if-loops please.

system/modules/isotope/library/Isotope/Frontend.php
```diff
@@ -630,36 +630,39 @@ public static function getPagesInCurrentRoot(array $arrPages, $objMember = null)
...
+        if (!($alias = Input::getAutoItem('product', false, true))
+            || ($objProduct = Product::findAvailableByIdOrAlias($alias)) === null
+        ) {
+            return $arrItems;
+        }
+
+        /** @var \PageModel $objPage */
+        global $objPage;
+
+        if ($objPage->type === 'error_404') {
```
aschempp: we should do this check before the *expensive* search for product in the database.

qzminski: Good point, just changed.

aschempp:
Should we do the same for 403 page?

qzminski:
I don't think so.

baumisimon:
Now the Product never shows in the breadcrumb:
line 636 of Frontend.php
```
if (!($objPage->type instanceof \PageRegular)
```
always returns true ($objPage->type isn't an instance of PageRegular).
If I replace this line with
```
if (!($objPage->type == 'regular')
```
it works as expected for me.

qzminski:
True, it should be

```
if (!($objPage instanceof \PageRegular)
```

aschempp:
Fixed in f13e7a41e94b6ee909113b4dd007fbb497434c94

Next time please open a new issue if something is closed and released so it can be correctly assigned to a milestone.

aschempp:
> True, it should be
> 
> ```
> if (!($objPage instanceof \PageRegular)
> ```

Fun fact: `$objPage` is a `PageModel`, never `PageRegular` in my setup 🤷‍♂

PR #1692, opened 2016-08-12
Fixed the changelanguage event listener had reference to wrong Isotop…

qzminski:
…e global value


aschempp:
We should probably also check if the array exists anyway?

qzminski:
Yup just added.

PR #1701, opened 2016-09-03
Fixed limiting the page picker if the product category filter is active

qzminski:
**How to reproduce**
1. Limit the backend ```Product management``` view to certain page (via ```Pages``` button in top left corner)
2. Edit the product
3. Click the ```Change selection``` button

**Problem**
The page tree in the picker is limited (```root```) to what have been chosen in the pages filter.

**Solution**
Do not run the breadcrumb function which sets the ```root``` of ```tl_page``` if in the popup.


aschempp:
So this change only works because `table` is not set for the regular view? Shouldn't we better check for something like `TL_SCRIPT === 'page.php'` or `popup == 1`?

qzminski:
Yup, that's much better. I have adjusted the PR.

PR #1728, opened 2016-10-19
Feature: extended initial sortingfields

HenryLamorski:
it should be possible for the shopadmin to select the initial sortingfield in FE-Module "Product List" from all avail. attributes. (at least for me :) )


HenryLamorski:
wrong source branch, reopen soon. sorry.

PR #1758, opened 2017-01-26
Deliver $dc to \Checkbox::getAttributesFromDca when generating variant options

civ23:
It is necessary for a/my custom attribute which will use options_callback in \Widget:: getAttributesFromDca as last parameter

file: system/modules/isotope/library/Isotope/Backend/Product/VariantGenerator.php
line: 41
`$arrField = \CheckBox::getAttributesFromDca($GLOBALS['TL_DCA'][$table]['fields'][$attribute], $attribute, null, $attribute, $table, $dc);`


aschempp:
Fixed in d4f76d7345cf123c0dbab52e4b18482a064946b8

PR #1778, opened 2017-02-23
Added the product_price shipping method

qzminski:


system/modules/isotope/library/Isotope/Model/Shipping/ProductPrice.php
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public function isAvailable()
+    {
+        if (!parent::isAvailable()) {
+            return false;
+        }
+
+        /** @var ProductCollectionItem $item */
+        foreach (Isotope::getCart()->getItems() as $item) {
+            if (!in_array($this->attributeName, $item->getProduct()->getType()->getAttributes(), true)) {
```
aschempp: You need to check if `hasProduct` here

system/modules/isotope/library/Isotope/Model/Shipping/ProductPrice.php
```diff
@@ -0,0 +1,90 @@
...
+            $cartPrice = $item->quantity * $cartProduct->{$this->attributeName};
+
+            if ($cartPrice > $price) {
+                $price   = $cartPrice;
+                $product = $cartProduct;
+            }
+        }
+
+        if ($product === null) {
+            return 0;
```
aschempp: This case should never be possible. As far as I can understand, it would only happen if all products are exempted from shipping. This should be checked in the availability method.

system/modules/isotope/library/Isotope/Model/Shipping/ProductPrice.php
aschempp: Also, this does not check for variant attributes!

qzminski:
Improved in d73958e.

PR #1835, opened 2017-07-31
fix for #1833

fritzmg:
This will log exceptions like so:
```
[31-Jul-2017 13:55:14] Exception in post-sale request
…/vendor/codefog/contao-haste/library/Haste/Http/Response/Response.php(259): This exception has no message. Use $exception->getResponse() instead.
#0 …/vendor/isotope/isotope-core/system/modules/isotope/postsale.php(166): Haste\Http\Response\Response->send()
#1 …/vendor/isotope/isotope-core/system/modules/isotope/postsale.php(234): Isotope\PostSale->run()
#2 {main}
```
instead of just
```
[31-Jul-2017 13:55:14] Exception in post-sale request
#0 …/vendor/isotope/isotope-core/system/modules/isotope/postsale.php(166): Haste\Http\Response\Response->send()
#1 …/vendor/isotope/isotope-core/system/modules/isotope/postsale.php(234): Isotope\PostSale->run()
#2 {main}
```
See #1833

Issue #1833, opened 2017-07-31
Exception in post sale request does not log actual exception


fritzmg:
* Contao 4.4.2
* Isotope 2.4.3

I just tested a payment and the postsale request made a log entry saying 
```
Exception in post-sale request. See system/logs/isotope_postsale.log for details.
```
However, when I look into the log file, there is no visible exception there:
```
[31-Jul-2017 13:21:26] New request to https://[…]/system/modules/isotope/postsale.php?mod=pay&id=[…].

Headers: array (
  […],
)

$_GET: array (
  'mod' => 'pay',
  'id' => […],
)

$_POST: array (
  […]
)

Body:
[…]

[31-Jul-2017 13:21:26] Exception in post-sale request
#0 /[…]/vendor/isotope/isotope-core/system/modules/isotope/postsale.php(166): Haste\Http\Response\Response->send()
#1 /[…]/vendor/isotope/isotope-core/system/modules/isotope/postsale.php(240): Isotope\PostSale->run()
#2 {main}
```
Looking at the [log mechanic](https://github.com/isotope/core/blob/master/system/modules/isotope/postsale.php#L180-L186), it looks like it's omitting the actual error:
```php
log_message(
    sprintf(
        "Exception in post-sale request\n%s\n\n",
        $e->getTraceAsString()
    ),
    'isotope_postsale.log'
);
```
`getTraceAsString()` does not include the actual error. The function should probably be changed to:
```php
log_message(
    sprintf(
        "Exception in post-sale request\n%s\n%s\n\n",
        $e->getMessage(),
        $e->getTraceAsString()
    ),
    'isotope_postsale.log'
);
```
Also the `System::log` mechanic seems odd:
```php
\System::log(
    sprintf(
        'Exception in post-sale request. See system/logs/isotope_postsale.log for details.',
        $e->getFile(),
        $e->getLine(),
        $e->getMessage()
    ),
    __METHOD__,
    TL_ERROR
);
```
`sprintf` is used with 3 arguments (the error file, line and message), but there is no placeholder for them in the string.

aschempp:
see #1838

PR #1838, opened 2017-08-03
Output actual exception message in postsale

richardhj:
I thought, the actual exception message would be helpful for debugging.
Just now, I have the stack trace, but not the resulting exception message.

For example, this is the output in my postsale log:
```
[02-Aug-2017 22:02:30] Exception in post-sale request
#0 /opt/users/www/system/modules/core/library/Contao/Database.php(207): Contao\Database\Statement->query('SELECT id, name...')
#1 /opt/users/www/composer/vendor/codefog/contao-haste/library/Haste/Util/Format.php(149): Contao\Database->query('SELECT id, name...')
#2 /opt/users/www/composer/vendor/codefog/contao-haste/library/Haste/Util/Format.php(123): Haste\Util\Format::dcaValueFromArray(Array, NULL, NULL)
#3 /opt/users/www/system/modules/cfg_isotope_stock_management/src/Entry/AbstractEntry.php(121): Haste\Util\Format::dcaValue('tl_iso_product', 'groups', NULL)
#4 /opt/users/www/system/modules/cfg_isotope_stock_management/src/Entry/AbstractEntry.php(97): Codefog\IsotopeStockManagement\Entry\AbstractEntry->sendOutOfStockNotification()
#5 /opt/users/www/system/modules/cfg_isotope_stock_management/src/Hooks.php(132): Codefog\IsotopeStockManagement\Entry\AbstractEntry->save()
#6 /opt/users/www/composer/vendor/isotope/isotope-core/system/modules/isotope/library/Isotope/Model/ProductCollection/Order.php(186): Codefog\IsotopeStockManagement\Hooks->postCheckout(Object(Isotope\Model\ProductCollection\Order), Array)
#7 /opt/users/www/composer/vendor/isotope/isotope-core/system/modules/isotope/library/Isotope/Model/Payment/Paypal.php(74): Isotope\Model\ProductCollection\Order->checkout()
#8 /opt/users/www/system/modules/isotope/postsale.php(178): Isotope\Model\Payment\Paypal->processPostsale(Object(Isotope\Model\ProductCollection\Order))
#9 /opt/users/www/system/modules/isotope/postsale.php(240): Isotope\PostSale->run()
#10 {main}
```

### DIFF
--- a/system/modules/isotope/postsale.php
+++ b/system/modules/isotope/postsale.php
@@ -167,24 +167,12 @@ class PostSale extends \Frontend
 
         } catch (\Exception $e) {
             \System::log(
-                sprintf(
-                    'Exception in post-sale request. See system/logs/isotope_postsale.log for details.',
-                    $e->getFile(),
-                    $e->getLine(),
-                    $e->getMessage()
-                ),
+                'Exception in post-sale request. See system/logs/isotope_postsale.log for details.',
                 __METHOD__,
                 TL_ERROR
             );
 
-            log_message(
-                sprintf(
-                    "Exception in post-sale request\n%s\n%s\n\n",
-                    $e->getMessage(),
-                    $e->getTraceAsString()
-                ),
-                'isotope_postsale.log'
-            );
+            log_message((string) $e, 'isotope_postsale.log');
 
             $objResponse = new Response('Internal Server Error', 500);
             $objResponse->send();
--- a/system/modules/isotope/postsale.php
+++ b/system/modules/isotope/postsale.php
@@ -179,7 +179,8 @@ class PostSale extends \Frontend
 
             log_message(
                 sprintf(
-                    "Exception in post-sale request\n%s\n\n",
+                    "Exception in post-sale request\n%s\n%s\n\n",
+                    $e->getMessage(),
                     $e->getTraceAsString()
                 ),
                 'isotope_postsale.log'


richardhj:
Duplicate of #1835, #1833

aschempp:
We should also remove the invalid `sprintf` arguments to the `System::log` call.

richardhj:
I think you should rather choose #1835 @aschempp ;-) This is already done there :-)

aschempp:
Yeah, but it's incorrectly adding file and line which is already present in the trace ;-)

richardhj:
RFC then @aschempp @fritzmg

fritzmg:
> Yeah, but it's incorrectly adding file and line which is already present in the trace ;-)

No, neither file nor line is present in the trace.

fritzmg:
Oh, you meant @richardhj PR, which simply converts the Exception to string, which includes the exception information _and_ the trace.

However @richardhj in your example output, the actual exception is still missing. Is that from before or after the change?

richardhj:
As stated in the [manual](http://php.net/manual/de/exception.tostring.php#refsect1-exception.tostring-examples), the exception including the line and file and the following stack trace are given.
I changed to string casting of the exception afterwards.

fritzmg:
Alight, was just confused/wondering :)

Issue #1855, opened 2017-08-24
Saferpay Payment Provider broken


virtualize:
https://github.com/isotope/core/blob/20b2425af49d166ec338b5e8c13b66af80700ea1/system/modules/isotope/library/Isotope/Model/Payment/Saferpay.php#L122

This condition will always fail as its returns true if the status code is not 200 OR the response message does not begin with "ERROR". 

So if we have a valid response (code = 200 and message has no error string), we break here and redirect to the failed page. WTF?

This bug is more than a year old, so I assume nobody ever tested this or is using this PSP. Unfortunately we do. Will prepare a PR...

aschempp:
That is not true. The second condition tests if the response **does** start with `ERROR:`.

virtualize:
Excuse me, but you're wrong here. You're last sentence describes what we want to achieve, but thats not what the expression does. Please see this simple test case:

````php
<?php

function test($intCode, $strResponse) {
    if ((int)$intCode !== 200 || 0 !== strpos($strResponse, 'ERROR:')) {
        return "Request CANCELED \n";
    }
    return "Request EXECUTED \n";
}

$testCases = [
    [400, 'ERROR: message'],
    [200, 'ERROR: message'],
    [400, 'SUCCESS: message'],
    [200, 'SUCCESS: message'],
];

foreach ($testCases as list($intCode, $strResponse)) {
    echo sprintf("Code: '%s', Message: '%s' => Result: %s",
        $intCode, $strResponse, test($intCode, $strResponse)
    );
}
````

micbis:
@aschempp  Hallo Andreas - wie schauts hier aus? Siehst du eine Chance dies zeitnah zu integrieren/fixen? Wir würden gerne wenn irgendwie möglich auf die offizielle Version setzen und nicht forken müssen. Können wir dich allenfalls unterstützen?

aschempp:
Closed in favor of #1857

aschempp:
Fixed in 4088a3bda1b53cf04437501fa99bc988dd2e861c

PR #1857, opened 2017-08-25
Fix Saferpay PSP URL check

virtualize:
Fix for #1855   
The check for a valid Saferpay redirect URL was broken. A valid URL was validated as false and so the payment process failed.

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/Payment/Saferpay.php
+++ b/system/modules/isotope/library/Isotope/Model/Payment/Saferpay.php
@@ -119,11 +119,16 @@ class Saferpay extends Postsale implements IsotopeOrderStatusAware
         $objRequest->setHeader('Content-Type', 'application/x-www-form-urlencoded');
         $objRequest->send(static::createPayInitURI, http_build_query($this->generatePaymentPostData($objOrder), null, '&'), 'POST');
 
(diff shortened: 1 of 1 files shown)


system/modules/isotope/library/Isotope/Model/Payment/Saferpay.php
```diff
@@ -119,13 +119,18 @@ public function checkoutForm(IsotopeProductCollection $objOrder, \Module $objMod
...
+        if ((int) $objRequest->code !== 200 || 0 === strpos($objRequest->response, 'ERROR:')) {
             \System::log(sprintf('Could not get the redirect URI from Saferpay. See log files for further details.'), __METHOD__, TL_ERROR);
             log_message(sprintf('Could not get the redirect URI from Saferpay. Response was: "%s".', $objRequest->response), 'isotope_saferpay.log');
 
             Checkout::redirectToStep('failed');
         }
 
+        // redirect immediately
+        if (\Contao\Validator::isUrl($objRequest->response)) {
+            \Controller::redirect($objRequest->response);
```
aschempp: Can you explain what this is intended for? If the result was not a URL, the header field would not make sense either?

aschempp:
I've fixed the URL check in 4088a3bda1b53cf04437501fa99bc988dd2e861c

Let's open a new issue if the redirect change is necessary.

Issue #1794, opened 2017-04-24
Fehlerhafter Pfad bei Produktbildern unter Contao 4.3


Ainschy:
Bilder die nicht angepasst werden müssen, erhalten den Pfad zum isotope Produktbild Ordner, dieser wird unter Contao 4 nicht richtig aufgelöst.

aschempp:
Das Problem ist vermutlich der fehlende Symlink in den `web`-Ordner...

mecintosch:
Kann den Fehler bestätigen. Im Backend wird das Produktbild angezeigt im Frontend nicht. Der Isotope Bildordner befindet sich unter Contao 4 außerhalb des /web Ordners.

Bleibt wohl nur die Bilder größer als gewünscht hochzuladen. 
Neu berechnete Bilder werden angezeigt.

aschempp:
@isotope/developers how should we solve this? The `contao:symlinks` command is currently not extendable. I can think of three solutions, but I'm not really happy with any of them:

 1. Use the `kernel.command` or `kernel.terminate` event to create the symlink when `contao:symlinks` is called. We can't tell the user about it though, at least not in the same manner as the Contao output.
 2. Add our own command. Which means the user has to call it, and we probably can't add it to the Managed Edition `initializeApplication` composer script.
 3. Move the folder to `/web` if the installation is Contao 4. Not sure how to handle existing installations though.

rabauss:
4. can't we move it to files/isotope? the files folder is symlinked by default from contao...
or why is the contao file manager not used for the pictures?

aschempp:
The solution will be to create a symlink, like the core does for all other files.

 1. Add a Contao Manager plugin that has the `ConfigPluginInterface`
 2. Use the config loading to register an event listener on `kernel.command`
 3. Add an event listener that adds the symlink when the Contao symlinks are created

fritzmg:
In the mean time you could just execute
```
ln -s ../isotope isotope
```
in within the `web/` folder.

aschempp:
Implemented in c551a9bdfa7f10e02a632fa52bf941d236b5fdfb

fritzmg:
The fix does not work under Windows. You'll get the following error message after a composer update:
```

[Symfony\Component\Filesystem\Exception\IOException]
Failed to create symbolic link from "..\isotope" to "C:\xampp\htdocs\website\app\..\web\isotope".
```
I'll probably create a PR for this on the weekend.

bytehead:
[...]

PR #1894, opened 2017-12-02
fix symlink command for Windows

fritzmg:
This PR actually fixes three things:

* Fixes a copy & paste error in the comments.
* Fixes an error if the `/isotope` folder was not already present before updating.
* Fixes the symlink generation under Windows by using `Contao\CoreBundle\Util\SymlinkUtil` instead.

See also https://github.com/isotope/core/issues/1794#issuecomment-347970250

### DIFF
--- a/system/modules/isotope/library/Isotope/ContaoManager/Plugin.php
+++ b/system/modules/isotope/library/Isotope/ContaoManager/Plugin.php
@@ -18,7 +18,7 @@ class Plugin implements ConfigPluginInterface
             $container->setDefinition(
                 'isotope.listener.console',
                 (new Definition('Isotope\EventListener\SymlinkCommandListener'))
-                    ->setArguments(['%kernel.project_dir%'])
+                    ->setArguments(['%kernel.root_dir%'])
                     ->addTag('kernel.event_listener', ['event' => 'console.terminate'])
             );
         });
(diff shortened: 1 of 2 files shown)


fritzmg:
Are there any objections to this PR? Having this bug for each Contao 4 + Isotope project (on Windows) is a little annyoing ;)

aschempp:
`project_dir` is only available in Contao 4.4. If the `SymlinkUtil` is available in previous Contao versions, can we still use `kernel.root_dir` instead?

fritzmg:
`SymlinkUtil` is available since Contao `4.0.3`: https://github.com/contao/core-bundle/commit/086caa2b1095db36eb924f42c369cd20db282d24

Hm, I am not sure, I will try again with `kernel.root_dir` and
```php
SymlinkUtil::symlink('../isotope', '../web/isotope', $this->rootDir);
```

fritzmg:
Previous to Contao `4.4` the core-bundle used
```php
$this->rootDir = dirname($this->getContainer()->getParameter('kernel.root_dir'));
```
to get the actual root dir, should be no problem to use that instead. I'll update the PR accordingly.

fritzmg:
PR updated. (Though I dislike the "double meaning" of the "root dir" variables.)

aschempp:
thanks!

PR #1932, opened 2018-04-22
Check for a non-null product type.

richardhj:
The product type can be null as of the API. See https://github.com/isotope/core/blob/master/system/modules/isotope/library/Isotope/Interfaces/IsotopeProduct.php#L46


aschempp:
Thanks! Should we check for `$objProduct instanceof IsotopeProduct` instead of `!== null`?

aschempp:
Perfect, thanks!

PR #1949, opened 2018-06-11
hotfix change AbstractButton to use own button template

zonky2:
hotfix change AbstractButton to use own button template e.g. bootstrap like

```
<?php
$icon         = '';
$iconTemplate = '<span class="oi oi-%s" aria-hidden="true" role="button"></span> ';

switch ($this->name) {
    case 'update':
        $icon = sprintf($iconTemplate, 'loop-circular');
        break;
    case 'add_to_cart':
        $icon = sprintf($iconTemplate, 'cart');
        break;
}
?>
<button type="submit" name="<?= $this->name ?>" class="submit <?= $this->name ?> <?= $this->classes ?> btn btn-primary btn-sm" value="<?= $this->label ?>"><?= $icon ?><?= $this->label ?></button>
```

### DIFF
--- a/system/modules/isotope/library/Isotope/Frontend/ProductAction/AbstractButton.php
+++ b/system/modules/isotope/library/Isotope/Frontend/ProductAction/AbstractButton.php
@@ -3,6 +3,7 @@
 namespace Isotope\Frontend\ProductAction;
 
 use Isotope\Interfaces\IsotopeProduct;
+use Isotope\Template;
 
 abstract class AbstractButton implements ProductActionInterface
 {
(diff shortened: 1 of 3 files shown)


Toflar:
I think this is an important one. Any reason why we should not change to `<button>` by default @aschempp ?

aschempp:
¯\_(ツ)_/¯

Isotope just always used `input` instead of `button`, I think historically because IE did not correctly handle `button`s. But I guess that's obsolete in the meantime… 😁

aschempp:
Added in 132fd00c4f22da7c29f69497b80d6d34ab3f1633

PR #1972, opened 2018-09-07
Please use releases instead of the dev-master

Petschko:
Same as here: https://github.com/dreipunktnull/DpnDHLBundle/pull/2

I would love, when you start to use my releases instead of the master-branch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         "contao/manager-plugin": "^2.0",
         "discordier/httprequestextended": "~1.1",
         "terminal42/contao-fineuploader": "^2.0",
-        "petschko/dhl-php-sdk": "dev-master@dev",
+        "petschko/dhl-php-sdk": "^0.3",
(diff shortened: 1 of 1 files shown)


aschempp:
Thanks a lot for your contributions! If I remember correctly, I used the master branch because there were no releases, or they were not updated to the latest changes. But I'm happily using the release version if available.

Is a `0.x` version intended for now? You're aware that `^0.3` will not install version 0.4, because 0-versions are allowed to break compatibility in every minor version?

Petschko:
I usually support backward-compatibility, even when I'm in Version <1.0

I mark old methods as deprecated and remove them some Versions later (It shows up in your PHP-Error log, when you use a deprected method)

To be honest I'm quite close to jump on Version 1.0, just need some more testing and add one method.

aschempp:
Please let me know when you're at `1.0` then please 😊

Petschko:
Alright, I will^^

PR #1989, opened 2018-10-16
[RTM] use Chosen JavaScript for multiple selects

fritzmg:
If you configure a product attribute of the type "select" and allow multiple selections, Isotope will currently display a regular `<select multiple>`:

screenshot_2018-10-16 produkte produkt id 13 bearbeiten - contao open source cms

These are rather cumbersome to use. However, Contao already comes with the "Chosen" JavaScript, which will make such large selects with multiple enabled a lot easer to handle:

untitled-2

The simple change in this PR would enable the "Chosen" JavaScript for such product attributes. I marked this PR as [RFC] - because now the `size` setting would be meaningless, because it only controls the height of the regular `<select multiple>`. Since the `size` setting is needed for other attribute types though (like file uploads) and it is within a subpalette of `multiple`, it cannot be easily removed just for the select attribute type.


aschempp:
Should we not rather add a checkbox to the attribute configuration so that option can be enabled?

fritzmg:
Hm... well if you want to make it configurable, you might as well also make it configurable whether it should be a select, radio or checkbox field.

aschempp:
What do you mean, you already select that in the attribute type?

fritzmg:
> What do you mean, you already select that in the attribute type?

Right 😁 

Updated the PR accordingly.

fritzmg:
btw. I should rebase this to `develop`, shouldn't I?

aschempp:
Yes it should be on `develop`. Also, if you simply name the field `chosen`, it will automatically "appear" in the `eval` array 😉

fritzmg:
Good point, I'll update the PR.

fritzmg:
PR updated and rebased!

aschempp:
Thanks!

PR #2002, opened 2018-12-07
Attribute options: Extended "foreignKey" format to support an additional sorting field

binron:
Attribute options get sorted by "label" (current behavior).

This change allows to specify an optional sorting field (default is still "label")

screen shot 2018-12-07 at 11 11 25


aschempp:
Did you test the sorting for label? Because we're using an alias, I don't think this will work?

PR #2009, opened 2019-01-14
use template for buttons

fritzmg:
Currently buttons of product actions are hardcoded to be `<input>`s. This PR would allow those buttons to be changed to `<button>`s for example, if desired, by overriding the `iso_button` template.


Toflar:
Somehow duplicate of https://github.com/isotope/core/pull/1949 ?

PR #2010, opened 2019-01-14
add product type css class

fritzmg:
Within the product type configuration you are able to define a CSS class:

screenshot_2019-01-14 shop-konfiguration - contao open source cms

According to the description, that CSS class is supposed to be added to the other product CSS classes in the product list and product reader. However, this does not actually work. The `cssClass` of the product type is actually not used anywhere currently.


aschempp:
Good catch! Fixed in 40d6cd220af49a6cb377c48fca4d9707748d4b02

PR #2038, opened 2019-05-01
fix tcpdf.php path

fritzmg:
The `tcpdf.php` path for the `contao/core-bundle` (for Contao 4.4) was missing a `src/`. I also added the path to the `tcpdf.php` for the new `contao/tcpdf-bundle`.

It did not matter anyway, since the `system/config/tcpdf.php` symlink should always exist.


aschempp:
I actually fixed this by manually requiring tcpdf for Isotope

PR #2053, opened 2019-07-01
bugfix currency converting

cubex-hro:
automator throws error when converting because url is redirected to https


aschempp:
Thanks!

PR #2054, opened 2019-07-04
Add the bank details fields to store config

qzminski:


aschempp:
Please move the fields after the address (both in palette and DCA fields). Also the labels should be similar, all other are "Please enter" and not "Here you can".

qzminski:
Done. I always use "Please enter" for mandatory fields and "Here you can" for optional though.

PR #2075, opened 2019-09-18
Introduce product collection log feature

qzminski:


aschempp:
@qzminski please review my latest changes if you have a minute 😊

qzminski:
Did you test it? I think `$objOldStatus` and `$objNewStatus` are already equal in `onsubmit_callback` there?

Otherwise if it works then I'm fine with it 🙂

aschempp:
> Did you test it? I think `$objOldStatus` and `$objNewStatus` are already equal in `onsubmit_callback` there?

You got me confused there for a minute 😆 
The old status is taken from the order, which has not been updated yet. `$dc->activeRecord` points to `tl_iso_product_collection_log`, not the order itself.

PR #2076, opened 2019-09-19
hasTax() Shortcut - helping with tax exemption

cubex-hro:
Order Class shortcut added to find out if an order has tax or not.


aschempp:
This could probably be added to the parent class, so we could check for taxes of a cart as well?

cubex-hro:
That would be useful.

aschempp:
Can you change the Pull Request accordingly?

cubex-hro:
done :)

Issue #2069, opened 2019-09-12
SQL errors when accessing stats


m-vo:
Accessing any of the stats in the backend an error like the following gets thrown (if you've got an sql server that has `sql_mode=only_full_group_by` configured):

`SQLSTATE[42000]: Syntax error or access violation: 1055 Expression #3 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'myschema.o.locked' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by`

Example query:

```SQL
SELECT c.id                                         AS config_id,
       c.currency,
       o.locked                                     AS date,
       COUNT(o.id)                                  AS total_orders,
       COUNT(i.id)                                  AS total_products,
       COUNT(DISTINCT o.id)                         AS total_orders,
       COUNT(DISTINCT i.id)                         AS total_products,
       SUM(i.quantity)                              AS total_items,
       SUM(i.tax_free_price * i.quantity)           AS total_sales,
       DATE_FORMAT(FROM_UNIXTIME(o.locked), '%Y%m') AS dateGroup
FROM tl_iso_product_collection o
         LEFT JOIN tl_iso_product_collection_item i ON o.id = i.pid
         LEFT JOIN tl_iso_orderstatus os ON os.id = o.order_status
         LEFT OUTER JOIN tl_iso_config c ON o.config_id = c.id
WHERE o.type = 'order'
  AND o.order_status > 0
  AND o.locked IS NOT NULL
GROUP BY config_id, dateGroup
HAVING dateGroup >= 201903
   AND dateGroup <= 201909
```

Here for example `date` cannot be omitted in the group by clause for instance. 

(maybe related #1886,  #2019)

m-vo:
Not yet, hope I'll get to this next week. 

But judging from the query it should be enough to add the missing column(s) to the `GROUP BY` block without affecting the result. I'm not sure where this query originates from / where it's put together.

aschempp:
Grouping by `locked` certainly can't be correct, that would group all orders with the same order date …

m-vo:
You're right. Could it be because it's used in `DATE_FORMAT(FROM_UNIXTIME(o.locked), '%Y%m') AS dateGroup` and that `dateGroup` is grouped (though this should'nt be a problem normally)? I think I would need to run this in a sql console and debug it...

m-vo:
closed in favor of #2077

PR #2077, opened 2019-09-19
Bugfix/full group by constraint

m-vo:
see #2069 

@aschempp I'm not sure which branch to target - please let me know if I should rebase.


aschempp:
master should be fine, I can then rebase it to a hotfix branch. Is the `date` field not used in the statistics?

m-vo:
The code below only uses `dateGroup` - and that's also the column that is grouped after.

aschempp:
Fixed in fd355eb15c76afeb702685ed83853aa750c927e0

PR #2085, opened 2019-10-01
Force creation of resized images to allow synchronous file access

m-vo:
Isotope triggers image resizing like this...

```php
$strImage = \Image::create($strFile, $size)->executeResize()->getResizedPath();
```

...but then, the returned path is directly used in file access calls like so:

https://github.com/isotope/core/blob/348a59524b3d599cdc0fb728515d3ad075cbe782/system/modules/isotope/library/Isotope/Model/Gallery/Standard.php#L385

This cannot work with deferred image resizing in Contao >= 4.8. The easiest fix would probably be to execute the resizing synchronously (see this PR). This is possible with https://github.com/contao/contao/pull/693 which was implemented to deal with the same situation in the Contao newsletter (where synchronous access is also neccessary).

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/Gallery/Standard.php
+++ b/system/modules/isotope/library/Isotope/Model/Gallery/Standard.php
@@ -358,11 +358,6 @@ class Standard extends Gallery implements IsotopeGallery
         try {
             $strImage = \Image::create($strFile, $size)->executeResize()->getResizedPath();
             $picture = \Picture::create($strFile, $size)->getTemplateData();
-
-            $objResizedFile = new File($strImage);
-            if(method_exists($objResizedFile, 'createIfDeferred')) {
-                $objResizedFile->createIfDeferred();
-            }
         } catch (\Exception $e) {
             \System::log('Image "' . $strFile . '" could not be processed: ' . $e->getMessage(), __METHOD__, TL_ERROR);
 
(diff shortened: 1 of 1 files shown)


aschempp:
is there no other way to get the image size of a deferred image?

m-vo:
Well, we could use something like ...
```php
$objResizedFile = \Image::create($strFile, $size)->executeResize();
$strImage = $objResizedFile->getResizedPath();
 // ...
$arrSize = [$objResizedFile->getTargetWidth(), $objResizedFile->getTargetHeight()];
```
... but there are more things like the watermark which would need to be adapted as well. I'm not sure if this is worth the effort.

m-vo:
I mean, I could move the call to resize synchronously down to the watermark part and use the solution from my last comment otherwise. Would you prefer this?

ausi:
> is there no other way to get the image size of a deferred image?

```php
$arrSize = (new Contao\File(rawurldecode($strImage)))->imageSize;
```

m-vo:
@aschempp I updated the PR to use `File->imageSize` where possible. 

@ausi Could you maybe also have a :eyes: at this solution? :smile:

m-vo:
Any news on this? :see_no_evil:

system/modules/isotope/library/Isotope/Model/Gallery/Standard.php
```diff
@@ -364,12 +365,19 @@ protected function getImageForType($strType, array $arrFile, $blnWatermark = tru
...
         }
 
+        $objResizedFile = new File(rawurldecode($strImage));
+
         // Watermark
         if ($blnWatermark
             && $this->{$strType . '_watermark_image'} != ''
             && ($objWatermark = \FilesModel::findByUuid($this->{$strType . '_watermark_image'})) !== null
         ) {
+            if(method_exists($objResizedFile, 'createIfDeferred')) {
```
aschempp: @ausi do we need to create the deferred size? I assume we don't need to if we just use `$objResizedFile->imageSize` to get the file size?

aschempp: Nevermind, I saw this was only the case when adding watermarks, where we obviously need the file 😆

PR #2088, opened 2019-10-24
Fixes some MySQL 8 Exceptions

Balthyr:
Got this error in the contao logs:
```
[2019-10-24 18:22:40] app.CRITICAL: An exception occurred. {"exception":"[object] (Doctrine\\DBAL\\Exception\\SyntaxErrorException(code: 0): An exception occurred while executing 'SELECT tl_iso_productcache.* FROM tl_iso_productcache WHERE uniqid='[anonymized-id]' AND (keywords='' OR keywords='') AND (expires>1571934160 OR expires=0) AND groups='' LIMIT 0,1':\n\nSQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'groups='' LIMIT 0,1' at line 1 at [anonymized]/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php:99, Doctrine\\DBAL\\Driver\\PDOException(code: 42000): SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'groups='' LIMIT 0,1' at line 1 at [anonymized]/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:109, PDOException(code: 42000): SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'groups='' LIMIT 0,1' at line 1 at [anonymized]/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:107)"} []
```
Maybe my modification is a "dirty hack" but it works with MySQL 8 and the website is loading without any errors.

PR #2093, opened 2019-11-12
Fix some of MySQL 8 compatibility issues

qzminski:


Toflar:
We could use `Database::quoteIdentifier()` here but quoting directly is okay for me too. I'll leave that up to @aschempp to decide.

fritzmg:
How does the core deal with that usually?

aschempp:
Instead of escaping, we should prefix the fields with the table name. That will also fix any potential issues when joining related table. See https://github.com/isotope/core/pull/2088/files for some more. @qzminski can you please update the PR?

qzminski:
Updated 👍🏻

Issue #1931, opened 2018-04-19
Contao 4 compatibility problem


koertho:
When using Isotope with Contao 4.4., I get following warnings: 

> Warning: Declaration of Isotope\Model\TypeAgent::createModelFromDbResult(Database\Result $objResult) should be compatible with Contao\Model::createModelFromDbResult(Contao\Database\Result $objResult)

> Warning: Declaration of Isotope\Model\TypeAgent::createCollectionFromDbResult(Database\Result $objResult, $strTable = NULL) should be compatible with Contao\Model::createCollectionFromDbResult(Contao\Database\Result $objResult, $strTable)

This should easily be fixed by using the contao namespaces.

Tested with:
- Contao 4.4.17 and Isotope 2.4.6
- Contao 4.4.18 and Isotope 2.5.0

Toflar:
No idea really.

giacomodefra:
The latest version of isotope is compatible with Contao 4 ?

aschempp:
Isotope is compatible with Contao 4. These are just warnings, they should not show up in production.

m-vo:
Using the old Contao global namespaces also makes working with tools like phpstan cumbersome. :disappointed:  As soon as I'm using an isotope model for instance, the analyzer will break (because the globally namespaced classes cannot be found with regular autoloading). 

They are all over the place though. I just did a rough search/replace and there were > 200 affected files. So not sure if this would ever be changed / merged in case of a PR.

aschempp:
If you use Composer, the Isotope files are being loaded by Composer. What's your problem exactly?

akroii:
I'm getting the same message in frontend and backend after I switched the debug mode to 'on'
Contao 4.8.5
Isotope 2.6.4

```
ErrorException:
Warning: Declaration of Isotope\Model\TypeAgent::createModelFromDbResult(Database\Result $objResult) should be compatible with Contao\Model::createModelFromDbResult(Contao\Database\Result $objResult)

  at vendor/isotope/isotope-core/system/modules/isotope/library/Isotope/Model/TypeAgent.php:365
[...]

PR #2103, opened 2020-01-10
fix type declaration error in Contao 4

fritzmg:
Fixes https://github.com/isotope/core/issues/1931

Not very pretty, but would get the job done.


fritzmg:
Wait, isn't it much easier to just use `\Contao\Database\Result` in the `TypeAgent` anyway? This would still be backwards compatible to Contao 3, would it not?

aschempp:
see d8e6df3aefb49b654de1b72304e55fa389446481

PR #2144, opened 2020-05-29
Force creation of Image

tsarma:
Contao 4.8 introduced a Deferred image resizing concept to boost performance. This may cause issue like no product image in Isotope Invoice PDF.

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/Gallery/Standard.php
+++ b/system/modules/isotope/library/Isotope/Model/Gallery/Standard.php
@@ -357,7 +357,18 @@ class Standard extends Gallery implements IsotopeGallery
 
         try {
             $strImage = \Image::create($strFile, $size)->executeResize()->getResizedPath();
-            $picture = \Picture::create($strFile, $size)->getTemplateData();
+
+            if (!is_file(TL_ROOT . '/' . $strImage)) {
+                $image = \System::getContainer()->get('contao.image.image_factory')->create(TL_ROOT . '/' . $strImage);
+                \System::getContainer()->get('contao.image.resizer')->resizeDeferredImage($image);
+            }
+
+            $picture = \Picture::create($strFile, $size)->getTemplateData();            
+
+            if (!is_file(TL_ROOT . '/' . $picture['img']['src'])) {
+                $src = \System::getContainer()->get('contao.image.image_factory')->create(TL_ROOT . '/' . $picture['img']['src']);
+                \System::getContainer()->get('contao.image.resizer')->resizeDeferredImage($src);
+            }
         } catch (\Exception $e) {
             \System::log('Image "' . $strFile . '" could not be processed: ' . $e->getMessage(), __METHOD__, TL_ERROR);
 


system/modules/isotope/library/Isotope/Model/Gallery/Standard.php
```diff
@@ -357,7 +357,18 @@ protected function getImageForType($strType, array $arrFile, $blnWatermark = tru
...
+                $image = \System::getContainer()->get('contao.image.image_factory')->create(TL_ROOT . '/' . $strImage);
+                \System::getContainer()->get('contao.image.resizer')->resizeDeferredImage($image);
+            }
+
+            $picture = \Picture::create($strFile, $size)->getTemplateData();            
+
+            if (!is_file(TL_ROOT . '/' . $picture['img']['src'])) {
+                $src = \System::getContainer()->get('contao.image.image_factory')->create(TL_ROOT . '/' . $picture['img']['src']);
+                \System::getContainer()->get('contao.image.resizer')->resizeDeferredImage($src);
+            }
```
fritzmg, line 371: What about images in a `srcset` or multiple `<source>`s? I mean, not that it really matters for PDF creation, but technically all the sources would neet to be resized in advance.

/cc @ausi

aschempp:
If I understand correctly, this would always create all images for all galleries of all products? So deferred images would essentially be _lost_ for Isotope?
Shouldn't we rather just create the images that are used in a document/PDF?

fritzmg:
True, it should only be done for the generation of images when creating a document.

system/modules/isotope/library/Isotope/Model/Gallery/Standard.php
tsarma: `srcset` is used only in browsers. There we don't have any problems. The problem is only in PDF generation.

tsarma:
>If I understand correctly, this would always create all images for all galleries of all products? So deferred images would essentially be lost for Isotope?

Yes. 

> Shouldn't we rather just create the images that are used in a document/PDF?

Ideally yes. There is at the moment no way of knowing from where the image is called.  As it is always called inside a template, which can be used by both PDF and website.

I think a possible solution might be to change method `generateMainImage()` signature.  `public function generateMainImage($createDeferredImage = false)`

fritzmg:
You could check for `TL_MODE`. When generating the PDF, it will be `BE`, otherwise `FE`, presumably.

tsarma:
> You could check for `TL_MODE`. When generating the PDF, it will be `BE`, otherwise `FE`, presumably.

It is so? Also for checkout process?

fritzmg:
Hm, you are right, that won't work. The dirty way would be to define a scope via `$GLOBALS` or something within the document generation …

ausi:
I think we should fix that when creating the PDF document in https://github.com/isotope/core/blob/master/system/modules/isotope/library/Isotope/Model/Document/Standard.php#L220

```php
if (method_exists(File::class, 'createIfDeferred')) {
    (new File(rawurldecode($args[2])))->createIfDeferred();
}
```

aschempp:
I have implemented @ausi's fix in a754933ab838799692d48de3dd274f57b10c4dab, @tsarma can you check if this works?

tsarma:
Yes, it works with @ausi's fix. Thank you

PR #2147, opened 2020-06-02
Update tl_iso_payment.xlf

tsarma:
The wording `Please enter your unique Saferpay account id` is a bit too vague and took me off. It may have been in an earlier version that Saferpay account Id has Terminal Id included. But now, in order to keep working with the current implementation of `/system/modules/isotope/library/Isotope/Model/Payment/Saferpay.php`, one has to provide both CustomerId and TerminalId, separated by `-`.

### DIFF
--- a/system/modules/isotope/languages/en/tl_iso_payment.xlf
+++ b/system/modules/isotope/languages/en/tl_iso_payment.xlf
@@ -303,10 +303,10 @@
         <source>Your project password for sofortüberweisung.de</source>
       </trans-unit>
       <trans-unit id="tl_iso_payment.saferpay_accountid.0">
-        <source>Saferpay Account-ID</source>
+        <source>Saferpay Account ID (CustomerId-TerminalId)</source>
       </trans-unit>
       <trans-unit id="tl_iso_payment.saferpay_accountid.1">
-        <source>Please enter your unique Saferpay account id.</source>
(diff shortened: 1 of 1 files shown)


system/modules/isotope/languages/en/tl_iso_payment.xlf
```diff
@@ -303,10 +303,10 @@
         <source>Your project password for sofortüberweisung.de</source>
       </trans-unit>
       <trans-unit id="tl_iso_payment.saferpay_accountid.0">
-        <source>Saferpay Account-ID</source>
+        <source>Saferpay Account ID (CustomerId-TerminalId)</source>
       </trans-unit>
       <trans-unit id="tl_iso_payment.saferpay_accountid.1">
-        <source>Please enter your unique Saferpay account id.</source>
+        <source>Please enter your unique Saferpay account id. It should be in the form CustomerId-CustomerId</source>
```
aschempp, line 309: Shouldn't it be `CustomerId-TerminalId` here as well?

tsarma: Oh yes, you are right it should be `CustomerId-TerminalId`. My mistake

aschempp: Please update the Pull Request then 😄

tsarma: Oh, I've updated it.

PR #2151, opened 2020-06-26
DHL Business Shipment: epk->ekp and invalid name of the function setEpk

mpitz:
There was a typo in the original module:

"epk" should be "ekp"

See also DHL Kundennummer (EKP):
https://www.dhl.de/content/dam/dhlde/external/dhl-anwenderhandbuch-gk-portal.pdf

The setEpk() funciton is deprecated:
https://github.com/Petschko/dhl-php-sdk/blob/451e5ffd03c17c639d841537d31d9ce26d05b79c/includes/Credentials.php#L226

There is a new setEkp() function:
https://github.com/Petschko/dhl-php-sdk/blob/451e5ffd03c17c639d841537d31d9ce26d05b79c/includes/Credentials.php#L217


Petschko:
See also: https://github.com/isotope/core/issues/2155

aschempp:
As the dependency is no longer supported, I doubt the method will ever be removed. Changing the name will only introduce a BC break / require a migration for existing users. I don't see any benefits currently, but thanks a lot for pointing it out!

PR #2152, opened 2020-06-26
Hotix adapt DCA to arrange widgets in iso_orders

zonky2:
Hotix adapt DCA to arrange widgets in iso_orders

https://easycaptures.com/fs/uploaded/1378/1820569673.png


aschempp:
Thank you @zonky2

PR #2161, opened 2020-08-28
Allow for response codes 303 and 307 for redirects

rhymeadam:
For Contao 4.10, Isotope needs to allow for response codes 303 and 307.  307 was used for a short time, but we should account for it anyway.

PR #2162, opened 2020-08-28
Allow for response codes 303 and 307 for redirects

rhymeadam:
For Contao 4.10, Isotope needs to allow for response codes 303 and 307.  307 was used for a short time, but we should account for it anyway.


aschempp:
Can you please adjust the non-minified file (as well) so we can see the actual changes?

aschempp:
closed in favor of https://github.com/isotope/core/pull/2161

PR #2194, opened 2021-02-17
Added `decodeEntities` to eval-Array of the Checkout-Step-Selection. 

### LINKED ISSUE #2193: Validierung schlägt fehl, wenn Bestellbedingungsformulare für Checkout angegeben werden
Da die generierten Optionvalues Backslashes enthalten, werden sie durch das Selectfield escaped. Dem Widget muss ein `['eval']['decodeEntities'] = true` mitgegeben werden, was fälschlicherweise dem MultiColumnWizard übergeben wird:
https://github.com/isotope/core/blob/e3b77e13b37b38429566d3e50318befc1561f989/system/modules/isotope/dca/tl_module.php#L327

Besser wäre es hier aufgehoben:
https://github.com/isotope/core/blob/e3b77e13b37b38429566d3e50318befc1561f989/system/modules/isotope/library/Isotope/Backend/Module/OrderConditionFields.php#L40

```php
'step' => [
                'label' => &$GLOBALS['TL_LANG']['tl_module']['iso_order_conditions']['step'],
                'inputType' => 'select',
                'options_callback' => ['Isotope\Backend\Module\OrderConditionFields', 'getSteps'],
                'eval' => [
                   'decodeEntities' => true,
                    'includeBlankOption' => true,
                    'style' => 'width:300px',
                    'columnPos' => 'checkout',
                ],
            ],
```

kollmeier:
Validation fails otherwise because of the backslashes in the option values. Closes #2193

### DIFF
--- a/system/modules/isotope/library/Isotope/Backend/Module/OrderConditionFields.php
+++ b/system/modules/isotope/library/Isotope/Backend/Module/OrderConditionFields.php
@@ -38,6 +38,7 @@ class OrderConditionFields extends \Backend
                 'inputType' => 'select',
                 'options_callback' => ['Isotope\Backend\Module\OrderConditionFields', 'getSteps'],
                 'eval' => [
+                    'decodeEntities' => true,
                     'includeBlankOption' => true,
                     'style' => 'width:300px',
                     'columnPos' => 'checkout',


aschempp:
Thanks for the PR! The option **labels** shouldn't be relevant for validation. Does the step **name** contain backslashes in your case? Did you create your own checkout steps?

kollmeier:
It is not the label, which causes the problems. The value (or name) of the options are generated and contain the FQDN of the checkout-step (containing backslashes).

Wether I create own checkout steps or not leads to the same result: The checkout step select field of the MultiColumnWizard (as mentioned in #2193) doesn't validate, because `\Isotope\CheckoutStep\PaymentMethod` gets compared to `&#92;Isotope&#92;CheckoutStep&#92;PaymentMethod` (for example)

After applying this submitted change, it works.

aschempp (approved):

aschempp:
Thank you @kollmeier

PR #2205, opened 2021-03-07
Check for order condition form

fritzmg:
By default, the `iso_order_conditions` multi column wizard widget produces an output like this, if you do not set any values:

```
array:1 [▼
  0 => array:5 [▼
    "form" => ""
    "step" => ""
    "position" => "before"
    "product_types" => ""
    "product_types_condition" => "oneAvailable"
  ]
]
```

This causes an invalid entry to be added to `$arrOrderConditions`. While this has no real effect in the front end, it seems safer and more optimised to ignore any order conditions entry where no actual form is configured (like the default one).

This also fixes an illegal array access that can occur, if the MCW saves wrong data into the field (which happened in one of our instances, though I cannot reproduce it).


aschempp (approved):

aschempp:
Thank you @fritzmg

PR #2206, opened 2021-03-10
Performance optimisation in backend product list

formatz:
As a user with restricted product types, we encountered a long page loading from more than 3 seconds (30 products per page) in Contao backend.
The products table contains more than 20'000 rows with lots of variants.

With this little change, the database request time is ~100ms (seems that FIND IN SET is not using the existing "pid" index).


aschempp (approved):

aschempp:
Thank you @formatz

formatz:
Thank you Andreas for your work

PR #2207, opened 2021-03-16
Bring back order editing functionality

qzminski:
This pull requests brings back previous order editing functionality, while maintaining the order log feature introduced in the last minor version.


fritzmg:
What is the problem with the current implementation in `2.7.0`?

aschempp:
> What is the problem with the current implementation in `2.7.0`?

There are multiple problems:
 1. it does not longer support "edit all" or "override all" to update the order status
 2. it no longer supports custom fields in the order detail (edit) view

aschempp (approved):

PR #2214, opened 2021-03-29
Remove system log error entry for missing tl_module.iso_notifications

fritzmg:
If you do not define a notification in the `iso_checkout` module then the following error will appear in the system log:

```
No notification for order ID …
```

This error entry is created here:

https://github.com/isotope/core/blob/5e3d7972cd91d782faf9f11a17deac89ccddc559/system/modules/isotope/library/Isotope/Model/ProductCollection/Order.php#L198-L222

Since this effectively makes this setting mandatory (otherwise your system log will be filled with errors with every order in the shop) it should be set to `mandatory` in the DCA as well to avoid confusion.

Alternatively

https://github.com/isotope/core/blob/5e3d7972cd91d782faf9f11a17deac89ccddc559/system/modules/isotope/library/Isotope/Model/ProductCollection/Order.php#L220-L222

could be removed altogether since it is perfectly possible to set the notification _only_ in the `OrderStatus`.

### DIFF
--- a/system/modules/isotope/dca/tl_module.php
+++ b/system/modules/isotope/dca/tl_module.php
@@ -685,7 +685,7 @@ $GLOBALS['TL_DCA']['tl_module']['fields']['iso_notifications'] = array
     'exclude'                   => true,
     'inputType'                 => 'select',
     'options_callback'          => array('NotificationCenter\tl_module', 'getNotificationChoices'),
-    'eval'                      => array('multiple'=>true, 'csv'=>',', 'includeBlankOption'=>true, 'chosen'=>true, 'tl_class'=>'w50'),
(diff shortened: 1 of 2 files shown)


fritzmg:
Personally I would prefer the latter as this gives more flexibility. But I am not sure what the original intention was.

aschempp:
I'm not sure this is a good idea, pretty sure there's a reason this is not mandatory. Most likely, don't we use the field in other modules as well? Also, using the order status is not ideal, as it might send notifications before the order is fully completed (therefore not containing all data).

fritzmg:
If it's not mandatory, then the error log entry should be removed.

fritzmg:
> Also, using the order status is not ideal, as it might send notifications before the order is fully completed (therefore not containing all data).

I don't think that's the case? All current core payment modules call `Order::checkout` and set their configured order status _afterwards_.

In any case I have adjusted the PR to remove the system log entry instead.

PR #2228, opened 2021-05-27
Fix mandatory file uploads in order conditions not working

fritzmg:
Currently mandatory file uploads in order conditions do not work. When the form is submitted, Isotope will stay at the `checkout/review` step, without showing any error. The problem is that when `checkout/process` processes the order conditions checkout step again, `$_FILES` will be empty at this point and thus the file upload widget will not validate and therefore the whole order condition step will not validate, even though it was valid during the `POST` request.

To fix this `$_FILES` will need to be repopulated for the cloned widget when the checkout is processed.

### DIFF
--- a/system/modules/isotope/library/Isotope/CheckoutStep/OrderConditions.php
+++ b/system/modules/isotope/library/Isotope/CheckoutStep/OrderConditions.php
@@ -114,7 +114,11 @@ class OrderConditions extends CheckoutStep implements IsotopeCheckoutStep
 
                 // Clone widget because otherwise we add errors to the original widget instance
                 $objClone = clone $this->objForm->getWidget($strField);
-                \Input::setPost($strField, $_SESSION['CHECKOUT_DATA'][$strField]);
+                if ($this->objForm->getWidget($strField) instanceof \uploadable) {
+                    $_FILES[$strField] = $_SESSION['FILES'][$strField];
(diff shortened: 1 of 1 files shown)


system/modules/isotope/library/Isotope/CheckoutStep/OrderConditions.php
```diff
@@ -114,7 +114,11 @@ function ($strName, &$arrDca) {
 
                 // Clone widget because otherwise we add errors to the original widget instance
                 $objClone = clone $this->objForm->getWidget($strField);
-                \Input::setPost($strField, $_SESSION['CHECKOUT_DATA'][$strField]);
+                if ($objClone instanceof \uploadable) {
+                    $_FILES[$strField] = $_SESSION['FILES'][$strField];
```
aschempp, line 118: doesn't this mean the file does not make it into checkout data as well?

fritzmg: checkout data should be fine:

https://github.com/isotope/core/blob/f04cab3ab2ecb5b78f4921cdd7eeb04655beca62/system/modules/isotope/library/Isotope/CheckoutStep/OrderConditions.php#L97-L107

aschempp (approved):

PR #2229, opened 2021-06-03
add suggest

zonky2:
add suggest isotope_quickproducts


fritzmg:
I don't think Isotope can or should maintain a list of Isotope extensions as suggests.

zonky2:
why not?

or is there an overview in the manual?

aschempp:
I agree with @fritzmg, we can't have a list of _every_ extension compatible with Isotope. The suggestions are (currently) things that are supported by the core. E.g. if you install some of the suggested payment API libraries, you will get the respective payment method because it is bundled with Isotope.

I wouldn't mind a list of well-known extensions in the manual, feel free to contribute that 😎 
https://github.com/isotope/docs

Issue #2174, opened 2020-11-14
sendNotificationMessage Hook ignored


janoschbatschi:
It seems that the Hook "sendNotificationMessage" is ignored if PayPal is selected as a payment module. When, e.g., Cash is selected the Hook gets called.
Contao: 4.4.54
Isotope: 2.6.14

aschempp:
there is no `sendNotificationMessage` hook in Isotope eCommerce?

janoschbatschi:
No, that's right. But due to the Hook being ignored only by the PayPal implementation which is part of Isotope eCommerce I thought this might be the right place. Using other payment methods the Hook gets called.

aschempp:
did you use PayPal Standard or PayPal PLUS?

janoschbatschi:
PayPal Standard

aschempp:
this should be fixed by https://github.com/isotope/core/pull/2120

PR #2231, opened 2021-06-16
Consider sendNotificationMessage Hook

c4y:
Since Isotope uses Notification-Center it should support sendNotificationMessage Hook from Notification-Center. With this Bundle you can send different notifications dependant on product: https://github.com/postyou/isotope_notification_product_extension
This does actually not work because postsale.php removes the Hook. See #2174

### DIFF
--- a/system/modules/isotope/postsale.php
+++ b/system/modules/isotope/postsale.php
@@ -235,6 +235,7 @@ class PostSale extends \Frontend
                     'getAttributesFromDca',
                     'loadDataContainer',
                     'replaceInsertTags',
+                    'sendNotificationMessage',
                 ]
             )
         );


fritzmg:
@aschempp why does the postsale entry point remove hooks at all? Shouldn't it only remove hooks that are known to cause issues for whatever reason?

aschempp:
@c4y thanks for the PR, I'll merge this in the next release.

@fritzmg because there were various issue with hooks due to the fact that not the full system is initialized/no actual page is there.

aschempp (approved):

bennyborn:
@aschempp While I welcome the addition of `sendNotificationMessage` to the whitelist I still think this might be the wrong way. The core itself comes with a manageable amount of hooks so why not just create a blacklist with the hooks known to cause trouble? I'm pretty certain there will be more problems using a whitelist approach in the future.

aschempp:
The postsale process should not be considered a safe environment for any hook we don't know we support it. There have been plenty of random issues with it and third-party extensions, and the current solution seems to work fine.

PR #2245, opened 2021-07-28
Also override language in request and translator

fritzmg:
Currently Isotope only overrides the language via `$GLOBALS['TL_LANGUAGE']`. However, this will cause the wrong translations to be shown in documents etc. if your templates use `<?= $this->trans('foobar', [], 'messages') ?>` rather than Contao translations.

This PR fixes that by doing all necessary steps that `PageRegular` does for example: https://github.com/contao/contao/blob/164a433f7ac3da55af2d01c2065d82e8ba4355eb/core-bundle/src/Resources/contao/pages/PageRegular.php#L63-L71

### DIFF
--- a/system/modules/isotope/library/Isotope/Frontend.php
+++ b/system/modules/isotope/library/Isotope/Frontend.php
@@ -843,7 +843,10 @@ class Frontend extends \Frontend
         if (class_exists(ContaoCoreBundle::class)) {
             /** @var ContainerInterface $container */
             $container = System::getContainer();
-            $container->get('request_stack')->getCurrentRequest()->setLocale($language);
+
+            if ($container->has('request_stack') && null !== ($request = $container->get('request_stack')->getCurrentRequest())) {
+                $request->setLocale($language);
+            }
 
             if ($container->has('translator')) {
                 $container->get('translator')->setLocale($language);
--- a/system/modules/isotope/library/Isotope/Frontend.php
+++ b/system/modules/isotope/library/Isotope/Frontend.php
@@ -853,6 +853,6 @@ class Frontend extends \Frontend
             }
         }
 
-        System::loadLanguageFile('default', $language);
+        System::loadLanguageFile('default', $language, true);
     }
 }
--- a/system/modules/isotope/library/Isotope/Frontend.php
+++ b/system/modules/isotope/library/Isotope/Frontend.php
@@ -853,6 +853,6 @@ class Frontend extends \Frontend
             }
         }
 
-        System::loadLanguageFile('default', $language, true);
+        System::loadLanguageFile('default', $language);
     }
 }
--- a/system/modules/isotope/library/Isotope/Frontend.php
+++ b/system/modules/isotope/library/Isotope/Frontend.php
@@ -12,6 +12,8 @@
 namespace Isotope;
 
 use Contao\Controller;
+use Contao\CoreBundle\ContaoCoreBundle;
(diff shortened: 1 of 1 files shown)


system/modules/isotope/library/Isotope/Frontend.php
```diff
@@ -828,4 +830,29 @@ public static function getIsotopeMessages()
...
+            if ($container->has('request_stack') && null !== ($request = $container->get('request_stack')->getCurrentRequest())) {
+                $request->setLocale($language);
+            }
+
+            if ($container->has('translator')) {
+                $container->get('translator')->setLocale($language);
+            }
+        }
+
+        System::loadLanguageFile('default', $language);
```
fritzmg: I have removed bypassing the static cache, I think that's not really necessary.

aschempp: I don't think that's correct. If you load english, then german and then want english again, you must use the `$blnNoCache` flag.

fritzmg: Why? The languages are stored in separate static caches.

aschempp: not sure what you mean, they are loaded in a global array? So if you load english then german and then english again, german would remain in the global array.

fritzmg: Yeah, I misunderstood the cache check in `System::loadLanguageFile`. I am now wondering though if the helper function should always force load the language, since that might degrade performance. I am not familiar on whether theses function are called in the front end multiple times as well?

aschempp: I think they are only loaded when printing a PDF or sending a notification. It _might_ be loaded multiple times, but then we also need to make sure the language is correct. Unfortunately, it seems we can't know the current language (to not load if not necessary)

fritzmg: > Unfortunately, it seems we can't know the current language (to not load if not necessary)

Hm, why not, I think we can do an early out with

```php
if ($GLOBALS['TL_LANGUAGE'] ?? null === $language) {
    return;
}
```

`$GLOBALS['TL_LANGUAGE']` should always be set to the actual language.

aschempp: `System::loadLanguageFile()` does not set `$GLOBALS['TL_LANGUAGE']`, so why would that work?

fritzmg: Within Isotope's own code base, `System::loadLanguageFile` with a language override is not executed without also setting the language in `$GLOBALS['TL_LANGUAGE']`. And if another extension does this, you would have wrong translations in other places for this request anyway, so I think it is fair to assume that the currently loaded language for translations is the same language as the one set for the current request.

aschempp: Maybe you can try to explain why you're trying to change the cache behaviour? It currently works and any change _might_ worsen it but not really improve anything?

fritzmg: The current PR fixes the issue of the translator not using the correct language and otherwise does not change the behaviour - except for the cache parameter for `System::loadLanguageFile`, which I will revert again. The current discussion is just about whether or not we can return early in the new helper method for performance reasons, but it's not really that important anyway.

aschempp (approved):

PR #2263, opened 2021-12-17
add XOF

MDevster:
add XOF to language files to allow usage in config https://github.com/isotope/core/blob/f22f6e3a27db8d870e8daeb6923b8930a9acee03/system/modules/isotope/config/currencies.php#L168

### DIFF
--- a/system/modules/isotope/config/config.php
+++ b/system/modules/isotope/config/config.php
@@ -451,7 +451,10 @@ $GLOBALS['ISO_NUM']["10000.00"]     = array(2, '.', "");
 $GLOBALS['ISO_NUM']["10,000.00"]    = array(2, '.', ",");
 $GLOBALS['ISO_NUM']["10.000,00"]    = array(2, ',', ".");
 $GLOBALS['ISO_NUM']["10'000.00"]    = array(2, '.', "'");
-
+$GLOBALS['ISO_NUM']["10000"] = [0, '.', ""];
+$GLOBALS['ISO_NUM']["10,000"] = [0, '.', ","];
+$GLOBALS['ISO_NUM']["10.000"] = [0, ',', "."];
+$GLOBALS['ISO_NUM']["10'000"] = [0, '.', "'"];
 
 /**
  * Hooks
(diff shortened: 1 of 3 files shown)


fritzmg:
Btw. there are libraries like [`rinvex/countries`](https://github.com/rinvex/countries) and others who can provide a list of countries and currencies. May be this could be used by Isotope, instead of managing them itself.

system/modules/isotope/config/config.php
```diff
@@ -451,7 +451,10 @@
 $GLOBALS['ISO_NUM']["10,000.00"]    = array(2, '.', ",");
 $GLOBALS['ISO_NUM']["10.000,00"]    = array(2, ',', ".");
 $GLOBALS['ISO_NUM']["10'000.00"]    = array(2, '.', "'");
-
+$GLOBALS['ISO_NUM']["10000"] = [0, '.', ""];
```
fritzmg: Have you tested this? I think I had an issue once with a definition like that due to it being interpreted as a numerical index.

system/modules/isotope/config/config.php
```diff
@@ -451,7 +451,10 @@
...
+$GLOBALS['ISO_NUM']["10,000"] = [0, '.', ","];
+$GLOBALS['ISO_NUM']["10.000"] = [0, ',', "."];
+$GLOBALS['ISO_NUM']["10'000"] = [0, '.', "'"];
```
fritzmg: The code style should be the same as above.

aschempp:
New currencies look good, but please remove the formatting option. It's not related to this PR, and theres a configuration for the number of digits anyway so it shouldn't be required.

MDevster:
@aschempp I think it does belong together. Our customer in Athopia wants a price formatting without decimal point, as also defined in the currencies "'XOF' => array('code' => 952, 'units' => 0)". But defining only the currency did not bring the desired success. For the field "Price formatting" also an adjustment had to be made so that the formatting fits. According to the customer, two decimal places were otherwise always displayed. Unfortunately, I do not directly manage the project and can not check this, because we only provided the theme.

aschempp:
Bildschirmfoto 2022-01-07 um 09 55 02

MDevster:
@aschempp Thanks for pointing this out, the customer had not made the adjustment. I have removed the changes in config.php. That's how it is when you only look in the code ;)

aschempp (approved):

PR #2266, opened 2022-02-03
Allow composer plugins

bytehead:


aschempp:
I just allowed them in my global composer setup, do we really need to add that to every package? 😂

bytehead:
🤷 I don't actually know what is the best solution. If somebody has not allowed it globally, it blocks the installation process.

aschempp:
I don't think it does. This only fixes the plugins if you checked out `isotope/core` locally and install its dependencies, right? It does not block anyone from using Isotope.

bytehead:
You're possibly right 😬

PR #2267, opened 2022-02-03
Use guzzle7 adapter instead of guzzle6

bytehead:
Be compatible with PHP 8, as guzzle6 is not officially: https://github.com/php-http/guzzle6-adapter/pull/78

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         "terminal42/contao-tablelookupwizard": "^3.2 || ^4.0",
         "terminal42/notification_center": "^1.0",
         "terminal42/dc_multilingual": "^4.0",
-        "guzzlehttp/guzzle": "^6.0",
+        "php-http/guzzle7-adapter": "^1.0",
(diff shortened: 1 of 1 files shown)


fritzmg:
May be it would be better to switch to `symfony/http-client` anyway for querying the PayPal API?

bytehead:
🤷 doesn't mind, if you use one of the HTTPlug compatible libraries, no?

fritzmg:
I don't know to be honest, I never really understood the concept and its benefits 😁

bytehead:
Or what is your motivation to go for `symfony/http-client`? :)

fritzmg:
Contao already requires `symfony/http-client` so if you need to make HTTP requests, why not use this instead. That's my thought at least.

bytehead:
Ah, didn't see that 🙈

bytehead:
Closing this in favor of a follow up `symfony/http-client` implementation.

bytehead:
See #2270

PR #2270, opened 2022-02-10
Replace guzzlehttp with symfony/http-client

bytehead:
Follow up for #2267.

Yet untested.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
         "tecnickcom/tcpdf": "^6.2.22",
         "patchwork/utf8": "^1.2",
         "symfony/filesystem": "^3.4 || ^4.0 || ^5.0 || ^6.0",
-        "symfony/http-client": "^4.4 || ^5.4"
+        "symfony/http-client": "^4.4 || ^5.4 || ^6.0"
     },
     "require-dev": {
         "contao/manager-plugin": "^2.0",
--- a/composer.json
+++ b/composer.json
@@ -43,10 +43,10 @@
         "terminal42/contao-tablelookupwizard": "^3.2 || ^4.0",
         "terminal42/notification_center": "^1.0",
         "terminal42/dc_multilingual": "^4.0",
-        "guzzlehttp/guzzle": "^6.0",
         "tecnickcom/tcpdf": "^6.2.22",
         "patchwork/utf8": "^1.2",
-        "symfony/filesystem": "^3.4 || ^4.0 || ^5.0 || ^6.0"
+        "symfony/filesystem": "^3.4 || ^4.0 || ^5.0 || ^6.0",
+        "symfony/http-client": "^4.4 || ^5.4"
     },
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -43,10 +43,10 @@
         "terminal42/contao-tablelookupwizard": "^3.2 || ^4.0",
         "terminal42/notification_center": "^1.0",
         "terminal42/dc_multilingual": "^4.0",
-        "guzzlehttp/guzzle": "^6.0",
         "tecnickcom/tcpdf": "^6.2.22",
         "patchwork/utf8": "^1.2",
-        "symfony/filesystem": "^3.4 || ^4.0 || ^5.0 || ^6.0"
+        "symfony/filesystem": "^3.4 || ^4.0 || ^5.0 || ^6.0",
+        "symfony/http-client": "^4.4 || ^5.4"
```
fritzmg: Can probably also be forward compatible with 6.x?

bytehead: Is it?

bytehead: Looks like: https://github.com/symfony/http-client/blob/6.0/CHANGELOG.md

bytehead: Added in 690e46941.

aschempp:
so we're only using Guzzle for PayPal stuff, right? I remember not using the official PayPal SDK back in the days because we didn't have composer. Not sure if we can/should change that instead of reimplementing the http lib 🤔 but that's probably a lot more work 😞

bytehead:
I thought about the same yesterday. There is for sure some SDK around from PayPal :-) I can look into it, need it anyways 😬

fritzmg:
Btw. if you use `->getContent(false)` you might not have to refactor so much regarding error handling.

bytehead:
I know, but I tried to implement some :-)

aschempp:
I have adopted the changes to PayPal standard in 1eff4627ea019c2deaaf21f37db64d75ad0e128f, but the PayPal API is still using Guzzle. Since I recently used a PayPal API lib I might be able to fix this 😊

aschempp:
Unfortunately, the PayPal API we are using is already deprecated. The new SDK uses a completely different API and would require a lot to be rewritten. Also, I don't like to require the PayPal SDK for _all_ users even if they don't require that payment method. Not sure how to fix that 😕

composer.json
```diff
@@ -97,5 +96,12 @@
                 "system/modules/isotope/config/upgrade.php"
             ]
         }
+    },
+    "config": {
+        "allow-plugins": {
+            "contao-components/installer": true,
+            "contao-community-alliance/composer-plugin": true,
+            "contao/manager-plugin": true
+        }
```
aschempp: please remove this from the PR

bytehead: removed in 9588b6f04.

bytehead:
yay ❤️

PR #2276, opened 2022-03-11
Start fixing warnings of undefined array key access

dmolineus:
This pull request is the initial start to fix warnings of undefined array key access when using isotope in PHP 8.

Disclaimer: This is not a complete fix just a result of trying to use PHP 8.1 and isotope 2.8.x-dev with Contao 4.13. It may be a starting point to generally fix the issues in Isotope.

### DIFF
--- a/system/modules/isotope/config/config.php
+++ b/system/modules/isotope/config/config.php
@@ -12,7 +12,7 @@
 /**
  * Backend modules
  */
-if (!isset($GLOBALS['BE_MOD']['isotope']) || !\is_array($GLOBALS['BE_MOD']['isotope']))
+if (!\is_array($GLOBALS['BE_MOD']['isotope'] ?? null))
 {
     array_insert($GLOBALS['BE_MOD'], 1, array('isotope' => array()));
 }
--- a/system/modules/isotope/config/config.php
+++ b/system/modules/isotope/config/config.php
@@ -12,7 +12,7 @@
 /**
  * Backend modules
  */
-if (!\is_array($GLOBALS['BE_MOD']['isotope']))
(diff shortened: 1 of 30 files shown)


system/modules/isotope/config/config.php
```diff
@@ -12,7 +12,7 @@
 /**
  * Backend modules
  */
-if (!\is_array($GLOBALS['BE_MOD']['isotope']))
+if (!isset($GLOBALS['BE_MOD']['isotope']) || !\is_array($GLOBALS['BE_MOD']['isotope']))
```
fritzmg:
```suggestion
if (!\is_array($GLOBALS['BE_MOD']['isotope'] ?? null))
```

system/modules/isotope/drivers/DC_ProductData.php
```diff
@@ -80,7 +80,7 @@ public function __construct($strTable)
         /** @var SymfonySession $objSession */
         $objSession = System::getContainer()->get('session');
         $arrClipboard = $objSession->get('CLIPBOARD');
-        if ($arrClipboard[$strTable]['mode'] === 'cutAll' && \Input::get('act') !== 'cutAll') {
+        if (isset($arrClipboard[$strTable]['mode']) && $arrClipboard[$strTable]['mode'] === 'cutAll' && \Input::get('act') !== 'cutAll') {
```
fritzmg:
```suggestion
        if ($arrClipboard[$strTable]['mode'] ?? null === 'cutAll' && \Input::get('act') !== 'cutAll') {
```

aschempp:
Can someone explain to me why the `deleteConfirm` should ever be empty? They are always defined in the fallback language?

fritzmg:
@aschempp you removed the changes for `$GLOBALS['TL_LANG']['tl_iso_attribute']['deleteConfirm']` etc., which will cause the problem again.

PR #2279, opened 2022-03-22
Do not preselect Document (Template) in Backend Order print.

tsarma:
Please insert a blank option in the document selection drop-down, when printing Orders from the Backend. In this way, the editor is forced to select the correct template.

### DIFF
--- a/system/modules/isotope/library/Isotope/Backend/ProductCollection/Callback.php
+++ b/system/modules/isotope/library/Isotope/Backend/ProductCollection/Callback.php
@@ -468,7 +468,7 @@ class Callback extends \Backend
             'label'      => &$GLOBALS['TL_LANG']['tl_iso_product_collection']['document_choice'],
             'inputType'  => 'select',
             'foreignKey' => 'tl_iso_document.name',
-            'eval'       => array('mandatory' => true),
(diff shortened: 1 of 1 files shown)


fritzmg:
I think this is something you should customise in your application and should not necessarily be changed for everyone.

tsarma:
@fritzmg Yes, I've customized the button callback in my installation as you have suggested. Just because I needed a blank option in the dropt-down, I had to copy over 50 lines of codes from Isotope's `Callback::printDocument`. This can be easily avoided by adding a blank option, I don't see any drawbacks or breaking of the system. Remember I am talking only about the function to print orders from the backend.

fritzmg:
Ah, I did not notice that this is not a DCA change.

aschempp:
The request seems valid to me. @fritzmg why would you think this should not be the default?

fritzmg:
It introduces an additional klick which would be unnecessary in our Isotope instances.

aschempp:
I have added a blank option in f620689256cb00bdfe89ae91cb2d4896518022b3 only if there is more than one option. @tsarma can you test that and make sure the mandatory selection is enforced?

tsarma:
@aschempp Thanks it works.

PR #2285, opened 2022-04-11
fix request cache and some warnings in the debug mode

MDevster:


system/modules/isotope/library/Isotope/BackendModule/BackendOverview.php
```diff
@@ -160,7 +163,7 @@ protected function getModule($module)
             // Include all excluded fields which are allowed for the current user
             if ($GLOBALS['TL_DCA'][$strTable]['fields']) {
                 foreach ($GLOBALS['TL_DCA'][$strTable]['fields'] as $k => $v) {
-                    if ($v['exclude'] && BackendUser::getInstance()->hasAccess($strTable . '::' . $k, 'alexf')) {
+                    if (isset($v['exclude']) && BackendUser::getInstance()->hasAccess($strTable . '::' . $k, 'alexf')) {
```
aschempp: this would also be to `true` if `$v['exclude'] = false`

system/modules/isotope/library/Isotope/BackendModule/InitializeListener.php
```diff
@@ -27,7 +27,7 @@ public function enableModuleTablesInSetup()
 
         foreach ($GLOBALS['ISO_MOD'] as $strGroup => $arrModules) {
             foreach ($arrModules as $strModule => $arrConfig) {
-                if (\is_array($arrConfig['tables'])) {
+                if (isset($arrConfig['tables']) && \is_array($arrConfig['tables'])) {
```
aschempp:
```suggestion
                if (\is_array($arrConfig['tables'] ?? null)) {
```

system/modules/isotope/library/Isotope/Model/Attribute.php
```diff
@@ -297,7 +297,7 @@ public function saveToDCA(array &$arrData)
                 $strGroup = '';
 
                 foreach ($arrOptions as $option) {
-                    if ($option['group']) {
+                    if (isset($option['group'])) {
```
aschempp:
```suggestion
                    if ($option['group'] ?? false) {
```

system/modules/isotope/library/Isotope/Module/ProductFilter.php
```diff
@@ -354,10 +354,13 @@ protected function generateSorting()
 
         if (0 !== \count($this->iso_sortingFields)) {
             $arrOptions = [];
+            $sortingDirection = null;
 
             // Cache new request value
             // @todo should support multiple sorting fields
-            [$sortingField, $sortingDirection] = explode(':', Input::post('sorting'));
+            if (strpos(Input::post('sorting'), ':') !== false) {
+                [$sortingField, $sortingDirection] = explode(':', Input::post('sorting'));
```
aschempp: `$sortingField` would no longer be set?

system/modules/isotope/library/Isotope/Model/RequestCache.php
```diff
@@ -414,7 +414,11 @@ public function saveNewConfiguration()
             return $this;
         }
 
-        $objCache = static::findOneBy(array('store_id=?', 'config=?'), $this->preSave(array((int) $this->store_id)));
+        $objCache = null;
+
+        if ($this->store_id != 0) {
```
aschempp, line 419: Can you please split this to a separate PR?

aschempp (changes requested):
Please always use `??` where possible and not `isset`.

Also, this is a bit confusing because it mixes a possible bugfix (Cache) with tons of CS fixes. Can you please split that into a separate PR?

aschempp:
I have incorporated all of the php8 fixes into the 2.8 branch. If you'd allow me push access to your repo, I could update this PR for you so it only contains the request cache changes 😊

MDevster:
@aschempp The access is granted

aschempp:
PR updated. Can you please elaborate on the change request? A store ID of `0` is totally valid, so I don't see why this should be changed?

aschempp:
ping @MDevster

MDevster:
@aschempp Unfortunately, I can no longer understand exactly because the changes arose from a support ticket and we just quickly helped the user that the store in Ethiopia is running again. His admin was probably not available. You can remove the change, he will contact us again after an update if necessary ;)

PR #2297, opened 2022-05-18
Do not use StringUtil::ampersand

fritzmg:
Isotope uses `StringUtil::ampersand` which is only available since Contao 4.9.22 (https://github.com/contao/contao/pull/3548).


aschempp:
I would rather revert to using the `ampersand()` function instead 🙈 
would you be willing to update the PR?

fritzmg:
Done 👍

aschempp (approved):
Thank you @fritzmg

PR #2291, opened 2022-05-01
Fix php8 array access errors

fritzmg:
This PR fixes some array access errors under PHP 8.

PR #2298, opened 2022-05-19
Xirdion/php8 access errors

Xirdion:
While using isotope with Contao 4.13 and PHP 8 I discovered some more array access errors in addition to the bugs that @fritzmg  has already fixed in #2291.

If you want, I can add my changes to that PR.

### DIFF
--- a/system/modules/isotope/dca/tl_iso_config.php
+++ b/system/modules/isotope/dca/tl_iso_config.php
@@ -351,7 +351,6 @@ $GLOBALS['TL_DCA']['tl_iso_config'] = array
                     'name' => array
                     (
                         'label'                 => ['&nbsp;'],
-                        'inputType'             => 'text',
                         'input_field_callback'  => array('Isotope\Backend\Config\AddressFieldsWizard', 'getNextName'),
                         'eval'                  => array('tl_class'=>'mcwUpdateFields'),
                     ),
--- a/system/modules/isotope/dca/tl_iso_config.php
+++ b/system/modules/isotope/dca/tl_iso_config.php
@@ -350,7 +350,7 @@ $GLOBALS['TL_DCA']['tl_iso_config'] = array
                 (
                     'name' => array
                     (
-                        'label'                 => ['&nbsp;', ''],
+                        'label'                 => ['&nbsp;'],
                         'inputType'             => 'text',
                         'input_field_callback'  => array('Isotope\Backend\Config\AddressFieldsWizard', 'getNextName'),
                         'eval'                  => array('tl_class'=>'mcwUpdateFields'),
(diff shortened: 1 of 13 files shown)


system/modules/isotope/dca/tl_iso_config.php
```diff
@@ -350,7 +350,8 @@
                 (
                     'name' => array
                     (
-                        'label'                 => ['&nbsp;'],
+                        'label'                 => ['&nbsp;', ''],
+                        'inputType'             => 'text',
```
aschempp: This does not look like it belongs here, but should be fixed in `multicolumnwizard`

system/modules/isotope/dca/tl_iso_config.php
```diff
@@ -648,7 +649,7 @@
         (
             'label'                 => &$GLOBALS['TL_LANG']['tl_iso_config']['newProductPeriod'],
             'exclude'               => true,
-            'default'               => array('unit'=>'days'),
+            'default'               => array('unit'=>'days', 'value'=>'1'),
```
aschempp: The default value should be empty

aschempp (changes requested):
Thanks for the PR! I have updated DC_ProductData to match the latest Contao 4.9/4.13, can you please rebase your PR and check what's still missing?

system/modules/isotope/dca/tl_iso_config.php
Xirdion: Yes, that's correct.
I was wondering if I should open a PR there or adjust the code here so that the error doesn't occur.
I have now reversed the change.

system/modules/isotope/dca/tl_iso_config.php
Xirdion: I have updated the file. thanks for the hint.

Xirdion:
I have rebased the PR. So only some changes are left in DC_ProductData.

system/modules/isotope/library/Isotope/Backend/ProductType/AttributeWizard.php
```diff
@@ -64,11 +64,13 @@ public function getColumns($objWidget)
         (
             'enabled'   => array
             (
+                'label' => '',
```
aschempp: I think these should be reverted as well since that should be fixed in MCW

system/modules/isotope/dca/tl_iso_config.php
```diff
@@ -351,6 +351,7 @@
                     'name' => array
                     (
                         'label'                 => ['&nbsp;'],
+                        'inputType'             => 'text',
```
aschempp: I think these should be reverted as well since that should be fixed in MCW

system/modules/isotope/library/Isotope/Backend/ProductType/AttributeWizard.php
```diff
@@ -64,11 +64,13 @@ public function getColumns($objWidget)
...
                 'inputType' => 'checkbox',
                 'eval'      => array('hideHead' => true),
             ),
             'name'      => array
             (
+                'label' => '',
```
aschempp: I think these should be reverted as well since that should be fixed in MCW

aschempp (changes requested):
Thanks a lot, this looks very good! 🤩 
Just some nitpicking about multicolumnwizard, then I'm happy to merge this

Xirdion:
Thanks for all the feedback.
I have removed the parts for multicolumnwizard.

aschempp (approved):

aschempp:
Thank you @Xirdion ❤️

PR #2304, opened 2022-06-10
Add specified category option to product list module

fritzmg:
In pretty much every instance where we use the Isotope shop we wanted to list products of a certain category in specific places in the front end. Currently the product list module does not allow to do that. This PR adds such an option:

_Note:_ I have based this against `2.8` as there is no `2.9` branch yet.

### DIFF
--- a/system/modules/isotope/dca/tl_module.php
+++ b/system/modules/isotope/dca/tl_module.php
@@ -18,6 +18,7 @@ $GLOBALS['TL_DCA']['tl_module']['palettes']['__selector__'][]               = 'i
 $GLOBALS['TL_DCA']['tl_module']['palettes']['__selector__'][]               = 'iso_emptyFilter';
 $GLOBALS['TL_DCA']['tl_module']['palettes']['__selector__'][]               = 'iso_tsdisplay';
 $GLOBALS['TL_DCA']['tl_module']['palettes']['__selector__'][]               = 'iso_tscheckout';
+$GLOBALS['TL_DCA']['tl_module']['palettes']['__selector__'][]               = 'iso_category_scope';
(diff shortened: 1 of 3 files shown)


fritzmg:
For anyone needing this feature right away:

```php
// contao/dca/tl_module.php
$GLOBALS['TL_DCA']['tl_module']['fields']['iso_category_scope']['options'][] = 'specific';
$GLOBALS['TL_DCA']['tl_module']['fields']['iso_category_scope']['eval']['submitOnChange'] = true;
$GLOBALS['TL_DCA']['tl_module']['fields']['iso_categories'] = [
    'exclude' => true,
    'inputType' => 'pageTree',
    'eval' => ['tl_class' => 'clr', 'mandatory' => true, 'multiple' => true, 'fieldType' => 'checkbox'],
    'sql' => ['type' => 'blob', 'length' => 65535, 'notnull' => false],
];
$GLOBALS['TL_DCA']['tl_module']['palettes']['__selector__'][] = 'iso_category_scope';
$GLOBALS['TL_DCA']['tl_module']['subpalettes']['iso_category_scope_specific'] = 'iso_categories';
```
```php
// src/EventListener/Isotope/FindCategories/SpecificCategoriesListener.php
namespace App\EventListener\Isotope\FindCategories;

use Contao\StringUtil;
use Isotope\Module\Module;

/**
 * Implements the "specific" product category list mode.
 */
class SpecificCategoriesListener
{
    public function __invoke(Module $module)
    {
        if ('specific' !== $module->iso_category_scope) {
            return false;
        }

        return StringUtil::deserialize($module->iso_categories, true);
    }
}
```
```php
// contao/config/config.php
use App\EventListener\Isotope\FindCategories\SpecificCategoriesListener;

$GLOBALS['ISO_HOOKS']['findCategories'][] = [SpecificCategoriesListener::class, '__invoke'];
```

aschempp:
Uhm, this should already possible by setting the reference page (like in the navigation module)?

fritzmg:
I see, so setting it to one of the "active category" modes and then define the reference page would work then.

Although with this you cannot filter the product list two two or more arbitrary categories, it's sufficient for our use case.

PR #2308, opened 2022-06-15
Automatically increase an attribute's database field length

fritzmg:
Currently if you have a select/radio attribute where the options are defined in the attribute itself, the database field that is created for `tl_iso_product` will have a hard coded length of `64`. However, this means that if you define an option with a value _larger_ than 64 characters, it will lead to an exception in the back end when trying to save the product with that option, as it won't fit into the database field.

This PR solves that by increasing the length of the database field (by multiples of 64) dynamically, according to the option's value lengths. For this to apply you will need to update the database yourself though after you have entered all the possible options for your attribute (via the Install Tool or `contao:migrate`).

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/Attribute/RadioButton.php
+++ b/system/modules/isotope/library/Isotope/Model/Attribute/RadioButton.php
@@ -11,6 +11,7 @@
 
 namespace Isotope\Model\Attribute;
 
+use Contao\StringUtil;
 use Isotope\Interfaces\IsotopeAttributeForVariants;
 
 /**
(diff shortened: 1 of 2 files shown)


aschempp:
why not use the actual length? Maybe we need to include the length of existing database values as well? Also, doesn't it update the database automatically if you change an attribute configuration?

fritzmg:
> why not use the actual length?

There is no actual length as the back end input field has no maximum length. Or do you mean the actual length of the string instead of multiple values of `64`?

> Maybe we need to include the length of existing database values as well?

Probably, yes. But it's at least an improvement to before.

> Also, doesn't it update the database automatically if you change an attribute configuration?

Yes, as I pointed out this would require you to update the database after you have changed the values.

aschempp:
>  Or do you mean the actual length of the string instead of multiple values of `64`?

yes

> > Also, doesn't it update the database automatically if you change an attribute configuration?
> 
> Yes, as I pointed out this would require you to update the database after you have changed the values.

uhm, why, if we update automatically?

fritzmg:
Does it already do that when you edit the options of an existing radio/select attribute? I didn't actually check 🙈

aschempp:
As far as I know it does, otherwise change if attribute types would not update the column.

PR #2321, opened 2022-07-12
Allow terminal42/contao-fineuploader v2 and v3

ameotoko:
I am not a pro with Composer, so please correct me if I step out of standards in `composer.json`.

Tested:

1. as a field in order conditions form, with multiple uploads allowed, moving to specified folder, and attaching uploaded files in NC notification via attachment tokens
2. as a frontend attribute of a product, with multiple uploads allowed, moving after upload, and renaming per Simple Tokens

Out of the box no breaking changes (if any) were encountered. If anoyone has any custom classes extending from FineUploader v2, those probably need to be rewritten though.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,7 @@
     "require-dev": {
         "contao/core-bundle": "4.9.*",
         "contao/manager-plugin": "^2.0",
-        "terminal42/contao-fineuploader": "^2.0",
+        "terminal42/contao-fineuploader": "^2.0 || ^3.0",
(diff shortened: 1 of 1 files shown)


fritzmg (approved):

fritzmg:
> If anoyone has any custom classes extending from FineUploader v2, those probably need to be rewritten though.

In that case they have required `terminal42/contao-fineuploader` in the correct version anyway, so no issue there.

aschempp (approved):

aschempp:
Thank you @ameotoko

PR #2339, opened 2022-07-23
Do not use `{{request_token}}` insert tag

ameotoko:
`{{request_token}}` insert tag is deprecated in Contao 4.13 and produces wrong value.

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/Product/Standard.php
+++ b/system/modules/isotope/library/Isotope/Model/Product/Standard.php
@@ -646,6 +646,9 @@ class Standard extends AbstractProduct implements WeightAggregate, IsotopeProduc
         $objTemplate->product_id = $this->getProductId();
         $objTemplate->module_id = $arrConfig['module']->id ?? null;
 
+        $strTokenName = System::getContainer()->getParameter('contao.csrf_token_name');
+        $objTemplate->requestToken = System::getContainer()->get('contao.csrf.token_manager')->getToken($strTokenName)->getValue();
+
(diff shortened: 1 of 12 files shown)


ameotoko:
I did not yet fixed `iso_collection_default.html5` and `iso_collection_favorites.html5` because I'm not sure which class is responsible for it.

aschempp:
Isotope (currently) still supports Contao 4.9, so we have to continue to use its features even if they are deprecated. But thanks for the PR, this will give us a head-start once we can fix this 😊

ameotoko:
This change does not conflict with Contao 4.9. Did you try it on Isotope Demo?

fritzmg:
> and produces wrong value.

The request token insert tag should not produce a wrong value, that would be a bug in Contao.

fritzmg (approved):
Other than that the changes lgtm and should be compatible with Contao 4.9 anyway.

ameotoko:
> The request token insert tag should not produce a wrong value, that would be a bug in Contao.

I think I meant that I was getting "CSRF token is invalid" message all the time. The value is not necessarily wrong.

fritzmg:
But if that happens only with the insert tag, it would still be a bug somewhere? Changing it to the service should make no difference.

ameotoko:
TBH I cannot reproduce it now either. May be the reason lied elsewhere, and I got A LOT fixed and updated in my Isotope app since this was opened. So probably you can disregard `produces wrong value` remark, but since the insert tag is deprecated anyway, merging this would still be a move forward, WDYT?

aschempp:
@ameotoko could we rebase this to 2.9 and use the `getDefaultTokenValue` then?

aschempp (approved):

aschempp:
Thank you @ameotoko

PR #2347, opened 2022-08-02
Handle case when host matches no roots, and no root with wildcard DNS exists

ameotoko:
If all root pages have `dns` set but request host does not match any of them, Contao throws `NotFoundHttpException`. However, if a product reader page is requested, then Isotope breaks this behaviour in PHP 8, failing with

```
TypeError:
current(): Argument #1 ($array) must be of type array, null given

  at vendor/isotope/isotope-core/system/modules/isotope/library/Isotope/Frontend.php:151
```

### DIFF
--- a/system/modules/isotope/library/Isotope/Frontend.php
+++ b/system/modules/isotope/library/Isotope/Frontend.php
@@ -145,6 +145,7 @@ class Frontend extends \Contao\Frontend
                 } elseif (isset($arrPages['*'])) {
                     $arrLangs = $arrPages['*']; // Empty domain
                 } else {
+                    // No domain match (see #2347)
                     return $arrFragments;
                 }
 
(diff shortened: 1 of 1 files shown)


system/modules/isotope/library/Isotope/Frontend.php
```diff
@@ -142,8 +142,10 @@ public function loadReaderPageFromUrl($arrFragments)
                 // Look for a root page whose domain name matches the host name
                 if (isset($arrPages[$strHost])) {
                     $arrLangs = $arrPages[$strHost];
-                } else {
+                } elseif (isset($arrPages['*'])) {
                     $arrLangs = $arrPages['*']; // Empty domain
+                } else {
+                    return $arrFragments;
```
fritzmg, line 149:
```suggestion
                    // No domain match (see #2347)
                    return $arrFragments;
```

fritzmg (approved):

aschempp (approved):

aschempp:
Thank you @ameotoko

PR #2349, opened 2022-08-06
Fix default translate value of images in MediaManager.php

### LINKED ISSUE #2348: Warning: Undefined array key "translate"
I noticed this migration recently – b6f8f98436080d38637bd049e5a2870b2b35bd59, which updates default translation field of all existing images from `''` to `none`. However, new images are still created with empty value by default. It seems that default value for this field is still missing, or got lost at some point while refactoring.

The sign of the problem is `Warning: Undefined array key "translate"`, when new products are being viewed in front end.

ameotoko:
Fixes #2348


aschempp (approved):

aschempp:
Thank you @ameotoko

PR #2359, opened 2022-08-19
Fix PHP8 warning

jamesdevine:
Contao 4.13.8, PHP 8.1.6. Creating a new store configuration results in the following error: Warning: Undefined array key "mandatory".

This PR fixes this by checking if the 'mandatory' key exists and defaults it to false if not.


aschempp:
Thank you @jamesdevine. I have slightly adjusted the PR because there seems to be a case of non-boolean values.

PR #2367, opened 2022-09-05
Fix 2.7.0, 2.7.1 and 2.8.0 migrations

fritzmg:
This fixes various issues with the 2.7.0, 2.7.1 and 2.8.0 migrations. Errors that can occur currently are:

* `Cannot execute queries while other unbuffered queries are active.`
* `Table 'tl_iso_product_collection_log' already exists.`
* `Field 'iso_readerMode' already exists.`

The former is fixed by calling `fetchAllAssoc` on the result, otherwise this error can occur in Contao 4.9.0+ because the result is buffered.

The latter two issues are fixed by setting `$blnNoCache` to `true` for the respective methods (`fieldExists` and `tableExists`). Within the 2.7.0 and 2.7.1 migrations this was probably the intention anyway - however the boolean was set for the wrong argument so far.

For the `2.8.0` migration fix I changed `Base::createDatabaseField` to also use `true` for the `$blnNoCache` parameter.

### DIFF
--- a/system/modules/isotope/library/Isotope/Upgrade/Base.php
+++ b/system/modules/isotope/library/Isotope/Upgrade/Base.php
@@ -29,11 +29,11 @@ abstract class Base extends System
      */
     protected function createDatabaseField($strField, $strTable)
     {
-        if (!Database::getInstance()->tableExists($strTable)) {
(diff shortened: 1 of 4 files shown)


aschempp (approved):

aschempp:
Thank you @fritzmg

PR #2370, opened 2022-09-15
hide empty fields

AlexanderWillner:
Steps that will reproduce the problem?
1. Order a product without providing a phone number
2. Send a mail notification about this order

What is the expected result?
```
...
12345 City
Country

mail@example.org
```

What happens instead?
```
...
12345 City
Country


mail@example.org
```

### DIFF
--- a/system/modules/isotope/languages/en/addresses.php
+++ b/system/modules/isotope/languages/en/addresses.php
@@ -23,8 +23,8 @@ $GLOBALS['ISO_ADR']['generic'] =
 ##hcard_country_name##</div>{endif}
 
 
-##hcard_tel##
-##hcard_email##</div>';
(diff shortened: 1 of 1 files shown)


aschempp:
Looks good! Did you test if this also removes the empty newlines at the end of the address (e.g. on a text mail)?

AlexanderWillner:
The use case is indeed text mails. 
Can you elaborate? The patch removes lines that would otherwise contain empty mail or phone details.

aschempp:
it looks like if you have an email but no phone, you would have three empty lines after the address?

AlexanderWillner:
After some internal tests, I've updated the patch accordingly.

aschempp:
thank you @AlexanderWillner

Issue #2375, opened 2022-09-21
Products with variants sorting by sku 


cboelter:
I've created a producttype with variants. So in the productlist is the product with its main variant shown. Everything fine.
If I enable the sorting by sku in the productlist the sorting is not working. I've checked the code a little bit and it seems the sorting is applied to the main product, the variant is not loaded before the sorting is applied, but the main product does not have an sku if the variants are enabled.

Is this a known limitation or is something wrong with my configuration?

I tested to load the default variant here: https://github.com/isotope/core/blob/2.8/system/modules/isotope/library/Isotope/Model/Product.php#L557-L559
That fixed the problem for me but I'm not sure about side effects.

aschempp:
I think that's currently a known limitation (if each variant has its own SKU). Also applies to any other field. The only solution I can think of would be a server-side sorting. There's already a similar implementation for prices, maybe they can be dynamically applied if the sorting attribute is configured on a variant.

aschempp:
Fixed by #2378

PR #2378, opened 2022-10-07
Add handling for sorting by variant attributes

cboelter:
As described here the sorting by variant attributes is not implemented yet: https://github.com/isotope/core/issues/2375
This pull request is a proposal to add sorting by variant attributes. I'm not really sure that this solves all cases, so it would be very nice to get some feedback :-)

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/Product.php
+++ b/system/modules/isotope/library/Isotope/Model/Product.php
@@ -546,25 +546,33 @@ abstract class Product extends TypeAgent implements IsotopeProduct
                         // will come before strings starting with a lowercase letter. To perform a case insensitive
                         // search, force the sorting order to be determined by a lowercase copy of the original value.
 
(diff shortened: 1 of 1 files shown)


aschempp:
Interesting approach. This only works if there is a default variant though, right?

cboelter:
You are right, I had the products without an default variant not in mind :-( ... I see there two approaches to handle them.

1. Ignore Products without default variants and use the main product
2. Get all variants and sort them first and use then the best matching variant for the sorting

Maybe you have more suggestions with your experience in isotope :-)

aschempp:
I have updated the PR to simplify things. If there **IS** a default variant, we will use your new approach which should work fine. If there is none, we simply continue with the old version (basically using an empty value). Are you ok with that?

cboelter:
Yes, I'm absolutely okay with that. Thanks! :-)

PR #2418, opened 2023-02-13
Fix the legacy registered hooks being overwritten by the annotation ones

qzminski:
If you have any hooks that are registered in the legacy way:

```php
// config/config.php
$GLOBALS['ISO_HOOKS']['generateProduct'][] = […];
```

They will get overwritten if there is at least one hook registered with annotations:

```php
use Isotope\ServiceAnnotation\IsotopeHook;

/**
 * @IsotopeHook("generateProduct")
 */
class MyListener {
    // …
}
```

This pull request aims to fix that.


aschempp:
@qzminski can you please check if everything still works as expected with my changes?

qzminski:
I can confirm it works as expected.

PR #2434, opened 2023-03-15
Fix php8 warnings: Undefined array key

michb:
fixed some php 8 warnings I ran into will testing. All warnings are "Undefined array key xxx"


aschempp (approved):

aschempp:
Thank you @michb!

PR #2457, opened 2023-05-22
Fix PHP 8 warning in order conditions form

ameotoko:
If order conditions form contains file upload field which is not mandatory, there will be a warning when the form is displayed:

```
ErrorException:
Warning: Undefined array key "FILES"

  at vendor/isotope/isotope-core/system/modules/isotope/library/Isotope/CheckoutStep/OrderConditions.php:123
  at Isotope\CheckoutStep\OrderConditions->generate()
     (vendor/isotope/isotope-core/system/modules/isotope/library/Isotope/Module/Checkout.php:325)
```

and later – when it is submitted:

```
ErrorException:
Warning: Undefined array key "FILES"

  at vendor/isotope/isotope-core/system/modules/isotope/library/Isotope/CheckoutStep/OrderConditions.php:105
```

### DIFF
--- a/system/modules/isotope/library/Isotope/CheckoutStep/OrderConditions.php
+++ b/system/modules/isotope/library/Isotope/CheckoutStep/OrderConditions.php
@@ -101,7 +101,7 @@ class OrderConditions extends CheckoutStep implements IsotopeCheckoutStep, Isoto
 
             $_SESSION['CHECKOUT_DATA'] = \is_array($_SESSION['CHECKOUT_DATA'] ?? null) ? $_SESSION['CHECKOUT_DATA'] : array();
             foreach (array_keys($this->objForm->getFormFields()) as $strField) {
-                if ($this->objForm->getWidget($strField) instanceof \uploadable) {
+                if ($this->objForm->getWidget($strField) instanceof \uploadable && !empty($_SESSION['FILES'][$strField])) {
                     $arrFile  = $_SESSION['FILES'][$strField];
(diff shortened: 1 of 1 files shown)


aschempp (changes requested):
Not sure that would be correct, because we're not threathing upload files as input fields if `$_SESSION['FILES']` is empty. I think we should use something like `$_SESSION['FILES'][$strField] ?? null` instead?

ameotoko:
You mean in the `if` condition?

```php
if ($this->objForm->getWidget($strField) instanceof \uploadable && ($_SESSION['FILES'][$strField] ?? null))
```

aschempp:
yeah, if the file is not uploaded, we shouldn't get the value from the form field, but just use an empty value. Casting (`??`) would do that I think?

aschempp:
Looks like you only changed one of the two places? 🙃

aschempp (approved):

aschempp:
Thank you @ameotoko

ameotoko:
I think my way is more concise. `!empty()` checks whether `$_SESSION['FILES'][$strField]` array key exists and has truthy value, which is exactly what we want. And if the condition is not satisfied, `$this->objForm->fetch($strField)` returns the proper default empty value for any widget, which can be `null` for `FormFileUpload` or empty array for `FineUploader`. Your version does not consider `FineUploader`.

aschempp:
The way Contao currently works, I don't think we're ever supposed to retrieve the value of an uploadable field? Something we recently discussed while updating fineuploader for Contao 5, unfortunately there's no "standard" in Contao for that.

PR #2461, opened 2023-06-09
Allow getGallery to use different Isotope Galleries

bright-cloud-studio:
We have a Product Reader page where we have multiple galleries being displayed. We wanted the ability to pick a different Isotope Gallery (Store configuration > Galleries) so I've added to getGallery a second parameter which is null by default but can be the ID of which gallery you want. Below that, if the ID is detected it replaces the current on inside $arrConfig.

I've tested this and it works perfectly. Feel free to change to any style you'd like, but for a three line change this adds very helpful functionality.

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/Product/Standard.php
+++ b/system/modules/isotope/library/Isotope/Model/Product/Standard.php
@@ -516,8 +516,9 @@ class Standard extends AbstractProduct implements WeightAggregate, IsotopeProduc
         $currentGallery          = null;
         $objTemplate->getGallery = function ($strAttribute, $galleryId=null) use ($objProduct, $arrConfig, &$currentGallery) {
             
-            if(!is_null($isoGalleryId))
-                $arrConfig['gallery'] = $isoGalleryId;
+            if(!is_null($isoGalleryId)) {
+                $arrConfig['gallery'] = $galleryId;
+            }
             
             if (null === $currentGallery
                 || $currentGallery->getName() !== $objProduct->getFormId() . '_' . $strAttribute
--- a/system/modules/isotope/library/Isotope/Model/Product/Standard.php
+++ b/system/modules/isotope/library/Isotope/Model/Product/Standard.php
@@ -514,8 +514,11 @@ class Standard extends AbstractProduct implements WeightAggregate, IsotopeProduc
 
         /** @var StandardGallery $currentGallery */
         $currentGallery          = null;
-        $objTemplate->getGallery = function ($strAttribute) use ($objProduct, $arrConfig, &$currentGallery) {
-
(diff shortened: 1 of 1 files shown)


system/modules/isotope/library/Isotope/Model/Product/Standard.php
```diff
@@ -514,8 +514,11 @@ public function generate(array $arrConfig)
 
         /** @var StandardGallery $currentGallery */
         $currentGallery          = null;
-        $objTemplate->getGallery = function ($strAttribute) use ($objProduct, $arrConfig, &$currentGallery) {
-
+        $objTemplate->getGallery = function ($strAttribute, $isoGalleryId=null) use ($objProduct, $arrConfig, &$currentGallery) {
```
aschempp:
```suggestion
        $objTemplate->getGallery = function ($strAttribute, $galleryId=null) use ($objProduct, $arrConfig, &$currentGallery) {
```

system/modules/isotope/library/Isotope/Model/Product/Standard.php
```diff
@@ -514,8 +514,11 @@ public function generate(array $arrConfig)
...
+            
+            if(!is_null($isoGalleryId))
+                $arrConfig['gallery'] = $isoGalleryId;
```
aschempp:
```suggestion
            if(!is_null($isoGalleryId)) {
                $arrConfig['gallery'] = $galleryId;
            }
```

aschempp (changes requested):
Please adjust the coding style. Also, will this work with the `$currentGallery` cache?

bright-cloud-studio:
I always tend to add iso to variable names when I am customizing something with Isotope but that doesn't make sense if I'm trying to add to the actual project. I've changed to the naming convention you suggested.

I can't say I am entirely sure what $currentGallery cache is, but we have been using this change on two different sites for a week or two now with no noticable issues.

aschempp:
> I can't say I am entirely sure what $currentGallery cache is, but we have been using this change on two different sites for a week or two now with no noticable issues.

I think this would only be a problem if you would try to use two different galleries on the same attribute in the same product (e.g. for the main image and for thumbnails). I'll look into this after merging the PR 😊

bright-cloud-studio:
So, another developer here was using my modification here and ran into the issue you predicted. This does not work when using the same image attribute for generateMainImage and generateGallery. Any suggestions on where to look next to adjust this? It looks like the getGallery function is called in ProductCollection.php in the addToTemplate function. Does that sound right?

aschempp:
I guess the easiest fix would be to remove the `$currentGallery` cache… or improve the cache to better handle configs.

aschempp (approved):

bright-cloud-studio:
I know in the grand scheme of things this is basically nothing, but I've never contributed to a piece of software I've used before. Pretty darn cool if you ask me.

aschempp:
Open Source FTW! ❤️

Issue #2468, opened 2023-06-24
No proper error handling in validateUpload of Isotope/Widget/MediaManager


chberger:
I'm facing an issue where I cannot upload product pictures. Even in the system log I can't find any hint pointing to the root cause. The only thing I'm getting is a ``CRITICAL: Uncaught PHP Exception ErrorException`` which tells me that the ``Isotope\Widget\MediaManager`` class tries to access an array at an undefined position:

```
[2023-06-24T11:36:47.450658+02:00] request.CRITICAL: Uncaught PHP Exception ErrorException: "Warning: Undefined array key 0" at /var/www/vhosts/#########/shop/vendor/isotope/isotope-core/system/modules/isotope/library/Isotope/Widget/MediaManager.php line 174 {"exception":"[object] (ErrorException(code: 0): Warning: Undefined array key 0 at /var/www/vhosts/#########//shop/vendor/isotope/isotope-core/system/modules/isotope/library/Isotope/Widget/MediaManager.php:174)"} []
```

The relevant implementation part looks like this:

```
        $varInput = '';

        try {
            $varInput = $objUploader->uploadTo($uploadFolder);
        } catch (\Exception $e) {
            $this->addError($e->getMessage());
        }

        if ($objUploader->hasError()) {
            $messages = System::getContainer()->get('session')->getFlashBag()->peek('contao_be_error');

            if (\is_array($messages)) {
                foreach ($messages as $strError) {
                    $this->addError($strError);
                }
            }
        }

        Message::reset();

        if (!\is_array($varInput) || empty($varInput)) {
            $this->addError($GLOBALS['TL_LANG']['MSC']['mmUnknownError']);
        }

        return $varInput[0];
    }

```

I guess in my case the ``$varInput`` isn't an array cause something goes wrong in `` $varInput = $objUploader->uploadTo($uploadFolder);``. Thus it shows that generic error message defined by ``$GLOBALS['TL_LANG']['MSC']['mmUnknownError']``. However, after that the code still tries to access the first position of ``$varInput`` which is obviously not working. 

Overall, I guess it would be helpful if that caught exception would be logged somehow. Otherwise I have no chance to identify the root cause at all.

Contao: 4.13.25
Isotope: 2.8.14
PHP:  8.1.19

chberger:
Ok, I've investigated it further. The ``Contao\Message``s are not being propagated properly, because this line is not working anymore:

[...]

PR #2472, opened 2023-06-30
MediaManager cannot handle errors if upload fails

ameotoko:
My Sentry reported this:

```
ErrorException: Warning: Undefined array key 0

#19 /vendor/isotope/isotope-core/system/modules/isotope/library/Isotope/Widget/MediaManager.php(174): Isotope\Widget\MediaManager::validateUpload
#18 /vendor/isotope/isotope-core/system/modules/isotope/library/Isotope/Widget/MediaManager.php(98): Isotope\Widget\MediaManager::ajaxUpload
#17 /vendor/isotope/isotope-core/system/modules/isotope/library/Isotope/Backend.php(335): Isotope\Backend::executePostActions
#16 /vendor/contao/core-bundle/src/Resources/contao/classes/Ajax.php(509): Contao\Ajax::executePostActionsHook
#15 /vendor/contao/core-bundle/src/Resources/contao/classes/Ajax.php(491): Contao\Ajax::executePostActions
#14 /vendor/contao/core-bundle/src/Resources/contao/classes/Backend.php(430): Contao\Backend::getBackendModule
#13 /vendor/contao/core-bundle/src/Resources/contao/controllers/BackendMain.php(168): Contao\BackendMain::run
#12 /vendor/contao/core-bundle/src/Controller/BackendController.php(49): Contao\CoreBundle\Controller\BackendController::mainAction
#11 /vendor/symfony/http-kernel/HttpKernel.php(163): Symfony\Component\HttpKernel\HttpKernel::handleRaw
#10 /vendor/symfony/http-kernel/HttpKernel.php(75): Symfony\Component\HttpKernel\HttpKernel::handle
#9 /vendor/symfony/http-kernel/Kernel.php(202): Symfony\Component\HttpKernel\Kernel::handle
#8 /vendor/symfony/http-kernel/HttpCache/SubRequestHandler.php(86): Symfony\Component\HttpKernel\HttpCache\SubRequestHandler::handle
#7 /vendor/symfony/http-kernel/HttpCache/HttpCache.php(481): Symfony\Component\HttpKernel\HttpCache\HttpCache::forward
#6 /vendor/symfony/framework-bundle/HttpCache/HttpCache.php(73): Symfony\Bundle\FrameworkBundle\HttpCache\HttpCache::forward
#5 /vendor/symfony/http-kernel/HttpCache/HttpCache.php(271): Symfony\Component\HttpKernel\HttpCache\HttpCache::pass
#4 /vendor/symfony/http-kernel/HttpCache/HttpCache.php(287): Symfony\Component\HttpKernel\HttpCache\HttpCache::invalidate
#3 /vendor/friendsofsymfony/http-cache/src/SymfonyCache/EventDispatchingHttpCache.php(126): Contao\ManagerBundle\HttpKernel\ContaoCache::invalidate
#2 /vendor/symfony/http-kernel/HttpCache/HttpCache.php(215): Symfony\Component\HttpKernel\HttpCache\HttpCache::handle
#1 /vendor/friendsofsymfony/http-cache/src/SymfonyCache/EventDispatchingHttpCache.php(98): Contao\ManagerBundle\HttpKernel\ContaoCache::handle
#0 /public/index.php(44): null
```


aschempp:
I'm not sure that's the right fix. There are multiple checks before that line that files are actually uploaded, including two lines above. So apparently the array does use another key than `0`? Can you reproduce the problem locally and check what the array key is?

ameotoko:
> So apparently the array does use another key than 0?

The bug is that the `$varInput` might not be an array at all. It is intialized as an empty string in line 150, and if `$objUploader->uploadTo()` throws an exception in the very next line, then `$varInput` remains a string and is never modified before that `return`.

https://github.com/isotope/core/blob/8a48dde2fdc571a5a7f4fd744860633fa2bf47aa/system/modules/isotope/library/Isotope/Widget/MediaManager.php#L150-L174

aschempp:
Did you get that upload error yourself? Can you maybe screenshot what's visible in the back end? In the code it looks like there would be at least 3 error messages for the same thing?

aschempp:
while you're at it, maybe you can also verify https://github.com/isotope/core/issues/2468 ?

ameotoko:
EDIT: #2468 precisely describes the problem below, but I only checked it after posting this comment😁

ORIGINAL: I have some new info related to this piece of code, but I am not sure whether it is related to this issue.

If filesize of uploaded image is bigger than allowed, than `$objUploader->hasError()` is `true`, but no exception is thrown and no flash messages are returned (`$messages` is empty). In the backend, "Upload failed" error message is shown without this PR's fix, and "The file could not be uploaded for unknown reason. Please check the system log." with the fix.

In the screenshot, first message is before the fix is applied, second - after:

image

The system log is empty btw, despite what the message says.

The reason is – there is no `contao_be_error` entry in the flash bag. Instead, it contains `contao.BE.error`, which contains the correct error message:

```diff
if ($objUploader->hasError()) {
-   $messages = System::getContainer()->get('session')->getFlashBag()->peek('contao_be_error');
+   $messages = System::getContainer()->get('session')->getFlashBag()->peek('contao.BE.error');

    if (\is_array($messages)) {
        foreach ($messages as $strError) {
            $this->addError($strError);
        }
    }
}
```

image

aschempp:
so fixing the name of the session bag name would make the change in this PR unnecessary?

ameotoko:
No, it would just fix my specific case and #2468. But as far as I understand, in PHP 8 you can't just initialize a variable as a string and then access it as an array later.

ameotoko:
This still needs to be fixed in `2.9.1`.

### To reproduce

While editing product in backend, upload an image with file size greater than allowed in `maxFileSize` in settings.

### Result

- Ajax request is terminated with server error `500` (see stack trace in [my comment at the top](https://github.com/isotope/core/pull/2472#issue-1782816955))
- The media manager widget shows generic error message `"Upload failed"`.

### Reason

There are two smaller problems in this issue:

1. Failure to provide proper error message to the user. This is described in #2468 and in [my comment above](https://github.com/isotope/core/pull/2472#issuecomment-1773483141)

2. Accessing a string variable as an array. `$varInput` is initialised as `string` in

https://github.com/isotope/core/blob/cbda2903bc25a8806468430edb8d96a3198e9751/system/modules/isotope/library/Isotope/Widget/MediaManager.php#L150

and if upload fails, `$varInput` is never reassigned and remains a `string`:
https://github.com/isotope/core/blob/cbda2903bc25a8806468430edb8d96a3198e9751/system/modules/isotope/library/Isotope/Widget/MediaManager.php#L152-L156

which is guaranteed to fail in

https://github.com/isotope/core/blob/cbda2903bc25a8806468430edb8d96a3198e9751/system/modules/isotope/library/Isotope/Widget/MediaManager.php#L174

### Fix

Fetch the error message from the correct storage key:

```diff
if ($objUploader->hasError()) {
-   $messages = System::getContainer()->get('session')->getFlashBag()->peek('contao_be_error');
+   $messages = System::getContainer()->get('session')->getFlashBag()->peek('contao.BE.error');

    if (\is_array($messages)) {
        foreach ($messages as $strError) {
            $this->addError($strError);
        }
    }
}
```

and make sure the request does not fail (otherwise the message is still not delivered):

```diff
- return $varInput[0];
+ return $varInput[0] ?? '';
```

aschempp (approved):

aschempp:
Thank you @ameotoko

PR #2482, opened 2023-08-07
Extend isotope dashboard  / overview page

error08:
Adds two new lines:
- sales current month
- sales current year
and a new column with the average value.

In addition, the total is now formatted with the currency of the associated configuration.

image


error08:
One problem: the queries check only for order_status>0 and do not consider a specific state e.g. completed. So it sums canceled orders, too. Can I use the field 'orderstatus_new' from config to filter the orders for ther summary? Or should I add a new one in the shop config?

aschempp:
but that was always the case, right? Maybe we could simply check for the payment date/"is paid" state of the order status? However new and unpaid orders would then likelynot be counted for the current day…

error08:
You‘re right we can use date_paid but loss the unpaid orders.

error08:
Another problem, according to the summaries, is that the totals are net values (tax_free_total) But also without a discounts. Is that correct? As a user, I would like to see my actual earnings (i.e. minus discount).

aschempp:
You're not earning what you give discount for, right? Shouldn't the total be that?

error08:
But the total is including the tax, isn't it?
I think we need the total without tax (tax_free_total) but minus the discount 😄

aschempp:
Hmm, it depends on whether that chart shows your profits or your revenue/sales… not sure if that's a one-size-fits-all sollution.

error08:
That's right! But in my option the net value (minus discount) or total with all surcharges and tax. The current mix make no sense for me. The same for the total sales report.

aschempp:
I'm happy with whatever you think makes most sense. Since I don't run my own shops, I can't really tell whats best (and I haven't used this in many years 😅).

error08:
So, I think now are all problems fixed.

aschempp:
Looks pretty good! I did some minor updates on the pull request. Be aware that german translations are managed on Transifex.com, so I removed the from the PR. 

One thing I noticed: can we change the summary to not be three separate collapsible sections, but all in one? Or would you think that does not make sense? Since the table columns are identical, maybe we can merge them into one? With the store config as a line to separate the items? Something like that?

Also, I think the first column title should be `Period` or `Time period` and not `Sales`, wdyt?

error08:
In my first attempt was everything in one table, but I think the most shops have only one/ a few configuration(s). So I think that's clearer this way. Perhaps the community comment that and we get more feedback according the configurations, wdyt?
Can you fix the first column. I'm now one week in holiday. I can fix that at the earliest in one week.

error08:
First column name is fixed. Ready to merge?

aschempp:
I have updated the table, I hope you're ok with that view?

Bildschirmfoto 2023-11-22 um 08 37 03

error08:
Yes, the presentation is good too. But now the currency format is wrong, € sign must be on the right side of the amount.

aschempp:
that's most likely because I configured my store configuration this way 🙃

aschempp:
Thank you @error08

PR #2483, opened 2023-08-07
switch first report member guests to apexcharts lib

error08:
My suggestion to replace the outdated xchart library with the modern apexcharts library.

With this PR I only changed the report MemberGuests.

Next I would change the SalesTotal, then the D3 Lib and the xcharts can be removed.

If this fits, the next PR will follow soon.
In addition, I plan to include the conversation rate in sales.


aschempp:
Great enhancement, thanks a lot! Can we update the `initializeChart` method instead of transforming the data for apexcharts? And could you rebase the pull request on Isotope 2.9 (or I can do that)?

Also I would prefer to remove the unnecessary plugin files and fix some code style. Let me know if I should apply that or you will.

error08:
Sure! I will refactor the initializeChart function and rebase the PR to 2.9. When I have trouble with the rebasing, I will contact you.

error08:
I had no idea how to more simplify the initializeChart function. For grouping the data, the time range is relevant. Later for the chart generation it is no longer necessary.

error08:
_"Also I would prefer to remove the unnecessary plugin files and fix some code style. Let me know if I should apply that or you will."_

Yes sure! But with my next PR regarding sales_total. At the moment sales_total still requires the old libraries.

aschempp:
Thanks a lot! I did mean to remove unnecessary files from the `apexcharts` folder, since we only seem to use one JS file (and probably some CSS), not all of the AMD stuff or the locales. Please also remove the export menu at the top, I don't think that makes sense on the chart (and it would mean we _really_ don't need locales 😅 )

error08:
I not sure if the lib works correct without the locale files. In my tests it seems it works.

system/modules/isotope_reports/library/Isotope/Report/MembersGuests.php
```diff
@@ -91,18 +91,26 @@ protected function compile()
...
             } else {
                 $arrDataGuests = $this->fillData($arrDataGuests, $objData);
                 // Generate chart data
-                $arrChart[$objData->currency . '_Guests']['data'][$objData->dateGroup]['y'] = ((float) $arrChart[$objData->currency . '_Guests']['data'][$objData->dateGroup]['y'] + $objData->total_sales);
+                $arrChart[$objData->currency . '_Guests']['data'][$objData->dateGroup][1] = ((float) $arrChart[$objData->currency . '_Guests']['data'][$objData->dateGroup][1] + $objData->total_sales);
             }
         }
 
+
+
```
aschempp: Please remove these empty line changes

system/modules/isotope_reports/templates/iso_report_members_guests.html5
```diff
@@ -1,100 +1,117 @@
...
 </div>
 
 <h2 class="sub_headline"><?= $this->headline ?></h2>
 
 <?= \Contao\Message::generate(true) ?>
 
 <div class="tl_listing_container list_view mod_iso_reports <?= $this->class ?>">
+  <?php $GLOBALS['TL_CSS'][] = 'system/modules/isotope_reports/assets/apexcharts/apexcharts.css'; ?>
+  <figure style="width:100%; height:300px" id="report-data"></figure>
+  <script src="<?= \Haste\Util\Debug::uncompressedFile('system/modules/isotope_reports/assets/apexcharts/apexcharts.min.js') ?>"></script>
```
aschempp: I think we can safely stop using `Debug::uncompressedFile` here, the feature is not included in Haste 5 anyway 😅

system/modules/isotope_reports/templates/iso_report_members_guests.html5
```diff
@@ -1,100 +1,117 @@
...
+          height: 350,
+          type: 'line',
+          zoom: {
+            enabled: false
+          },
+        },
+        series: <?= json_encode(array_values($this->chart)) ?>,
+      yaxis: {
+        title: {
+          text: <?= '"' . $GLOBALS['TL_LANG']['ISO_REPORT']['sales#'] . '"' ?>,
```
aschempp:
```suggestion
          text: "<?= $GLOBALS['TL_LANG']['ISO_REPORT']['sales#'] ?>",
```

aschempp (changes requested):
Looks very good! Just some minor nitpicks then I think we can merge this 🎉

error08:
Please wait for merging the PR, I found an error in the weekly and daily view. Here is a difference between the chart and table. The bug seems to have existed for a long time.

Because of that line 

` $intStart   = strtotime('first day of this month', $intStart);`

in the function `initializeChart` the chart and the table starts with different time range. (e.g. day or week)
Is there a specific reason for?

aschempp:
> in the function `initializeChart` the chart and the table starts with different time range. (e.g. day or week) Is there a specific reason for?

I honestly have no idea, it has been at least 5 years since I touched that 😂 It's not about monday or sunday start of the week, right?

error08:
I removed it. Now I think the PR is complete.

aschempp (approved):

aschempp:
Thank you @error08 !

PR #2487, opened 2023-10-07
Adaption-01-for-C5

Ernestopheles:
1. adaption to make Isotope compatible to Contao 5
Exchanged deprecated function call array_insert in config.php

Tested with cypress, testcases see https://github.com/Ernestopheles/isotopeDemo/tree/cypress


Ernestopheles:
accidentially added to 2.8

PR #2490, opened 2023-10-13
Prevent "class not available" error (Adaption towards Contao 5 Step 002)

Ernestopheles:

### DIFF
--- a/system/modules/isotope/library/Isotope/Interfaces/IsotopeProductCollection.php
+++ b/system/modules/isotope/library/Isotope/Interfaces/IsotopeProductCollection.php
@@ -16,7 +16,7 @@ use Haste\Units\Mass\Scale;
 use Isotope\Model\Config;
 use Isotope\Model\ProductCollectionItem;
 use Isotope\Model\ProductCollectionSurcharge;
-use Isotope\Template;
+use Contao\Template;
 
 
 /**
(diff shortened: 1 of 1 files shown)


Ernestopheles:
First adaption (002) towards Contao 5.

Take Template from Isotope\Template instead of \Template to prevent "class not available" error in ProductCollection::addToTemplate.

This error occured during cache warm-up under Contao 5.2, PHP 8.1 as class \Template was not available. 

system/modules/isotope/library/Isotope/Interfaces/IsotopeProductCollection.php
```diff
@@ -16,6 +16,7 @@
 use Isotope\Model\Config;
 use Isotope\Model\ProductCollectionItem;
 use Isotope\Model\ProductCollectionSurcharge;
+use Isotope\Template;
```
aschempp: This should be `Contao\Template`, it does not have to be an Isotope-specific one.

aschempp (changes requested):
Good find! And a perfect example of something we can & should & will fix in the 2.9 version since that affects Contao 4 as well. Just one minor change request regarding the namespace.

system/modules/isotope/library/Isotope/Interfaces/IsotopeProductCollection.php
Ernestopheles: Done

aschempp (approved):

aschempp:
Thank you @Ernestopheles

PR #2492, opened 2023-10-14
new report - member registrations

error08:
Add new report member registrations using new apexcharts library.


aschempp:
Unfortunately, I get an error message when trying this locally:

```
An exception occurred while executing a query: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '(ORDER BY m.dateAdded) as all_members,
DATE_FORMAT(FROM_UNIXTIME' at line 3
```

Since I'm running MySQL 5.7 (MAMP on macOS), it could be that `ORDER BY` inside the SELECT is not supported on older databases?

error08:
Yes you are right!  The window functions (sum in that case) are new in Version 8. But 5.7 has reached EOL. Should we support that version futhermore? How do you handle the base requirements for Isotope? Can they only be updated with a new major version of Isotope?

aschempp:
Well we cannot require a minimum version of MySQL (technically, on a Composer level). And Contao perfectly supports MySQL 5.7. So it would be a PITA to have a different dependency. Appart from the fact that my default hoster still has 5.7 (will be changed early next year) 😅

error08:
Ok, I see! I will try an alternative query.

error08:
I removed the window function and solved it via code. I think it is a good alternative. Now the code compatible to MySQL 5.x.

system/modules/isotope_reports/library/Isotope/Report/MembersRegistration.php
```diff
@@ -57,15 +56,27 @@ protected function compile(): void
...
             // Generate chart data
-            $arrChart['series'][0]['data'][$objData->dateGroup] = (int)$objData->all_members;
+            $arrChart['series'][0]['data'][$objData->dateGroup] = (int)$all_members;
             $arrChart['series'][1]['data'][$objData->dateGroup] = (int)$objData->new_members;
         }
 
+        $arrData['rows'] = array_reduce($arrData['rows'], function ($arrCarry, $arrItem) {
+            $prevItem = sizeof($arrCarry) > 0 ? $arrCarry[sizeof($arrCarry) - 1] : null;
+            if($arrItem['columns']['all_members']['value'] == 0 && $prevItem != null) {
+                echo "item:".$arrItem."|prev:".$prevItem;
```
aschempp: is this some leftover code?

aschempp:
I tried this PR locally. But (probably) because I did not have any registration/member data, I got an empty page and an error in apexchart
`Unhandled Promise Rejection: TypeError: null is not an object (evaluating 't.toString')`

I guess there's a problem somewhere if the data/row is empty?

error08:
Hi, I tried it without members / member data, but I cannot reproduce the error. I need more information about this.

aschempp:
Could you add a screenshot what this report is supposed to look like? Here's what I see. I guess the chart error is because some of the months do not have a new registration?

Bildschirmfoto 2024-02-05 um 10 49 44

error08:
Hi, here a screenshot

image

I have update the PR, please try it again. I see no such error.

image

error08:
@aschempp I think now is fine and ready to merge?

aschempp:
I'm sorry, I don't know what I'm doing wrong. I tried it again, but nothing changed. As you can see in the previous screenshot, I don't even have the same columns in the HTML table. Are you sure you pushed all changes?

error08:
I have found the problem. The language keys in the English file was wrong.

aschempp:
Thank you @error08! I fixed a minor issue with language collision.

PR #2494, opened 2023-10-14
Get rid of array_insert function (Adaption to Contao5 Step 005)

Ernestopheles:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,6 @@
         "hashids/hashids": "^4.1"
     },
     "require-dev": {
-        "contao/core-bundle": "4.13.*",
         "contao/manager-plugin": "^2.0",
         "terminal42/contao-fineuploader": "^2.0 || ^3.0",
         "terminal42/contao-changelanguage": "^3.0",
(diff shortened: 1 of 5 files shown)


Ernestopheles:
Next adaption (005) towards Contao 5.

Deprecated PHP function array_insert replaced by Contao\ArrayUtil::arrayInsert() method

composer.json
```diff
@@ -52,7 +52,7 @@
         "hashids/hashids": "^4.1"
     },
     "require-dev": {
-        "contao/core-bundle": "4.9.*",
+        "contao/core-bundle": "4.13.*",
```
aschempp: You can remove this line. It forced the dev-version of Isotope to install 4.9 instead of 4.13, because the requirement would allow for that. But since 4.13 is the last release, we don't need it anymore 🙃

aschempp (changes requested):
This will obvioiusly raise the minimum dependency of Isotope 2.9 to Contao 4.13, but I think this is a step we have to take to be able to update our code for Contao 5. Just one minor change request here.

composer.json
Ernestopheles: Thanks, done

aschempp (approved):

aschempp:
Thank you @Ernestopheles

PR #2495, opened 2023-10-15
Get rid of TL_MODE constant usage (Adaption to Contao5 Step 004)

Ernestopheles:


Ernestopheles:
Next adaption (004) towards Contao 5.

Replace the TL_MODE query with a Scopematcher check.

Added a Helper-class to enhance readability, maintenance, reusage and test.  

system/modules/isotope/library/Isotope/Helper/Scope.php
```diff
@@ -0,0 +1,54 @@
...
+     * @return bool
+     */
+    public static function isBackend(): bool
+    {
+        $container = System::getContainer();
+        $scopeMatcher = $container->get('contao.routing.scope_matcher');
+        $requestStack = $container->get('request_stack');
+
+
+        return $scopeMatcher->isBackendRequest($requestStack->getCurrentRequest() ?? Request::create(''));
```
aschempp: if no request exists, we should not generate one but return `false` (since it cannot be a backend request on command line for example).

system/modules/isotope/library/Isotope/Helper/Scope.php
```diff
@@ -0,0 +1,54 @@
...
+     * @return bool
+     */
+    public static function isFrontend(): bool
+    {
+        $container = System::getContainer();
+        $scopeMatcher = $container->get('contao.routing.scope_matcher');
+        $requestStack = $container->get('request_stack');
+
+
+        return $scopeMatcher->isFrontendRequest($requestStack->getCurrentRequest() ?? Request::create(''));
```
aschempp: if no request exists, we should not generate one but return `false` (since it cannot be a frontend request on command line for example).

aschempp (changes requested):
I generally like the helper class - or the idea of using something like this, because it simplifies a lot of code. However, I wonder if we should have a more generic class which can contain other methods (e.g. to get the project director instead of `TL_ROOT`).

Ernestopheles:
We can create more Helper classes. An advantage of having one class just for Scope matching is the easy readable use, we have "Scope::isBackend" instead of perhaps *Helper::isBackend" or "Helper::scopeIsBackend"

aschempp:
Yes, but since all those helper functions somewhat belong together, I would prefer to group them. Similar to the Contao `StringUtil` class. Maybe we could have a `\Isotope\CompatibilityHelper::isBackendScope()` and `\Isotope\CompatibilityHelper::getProjectDir()`?

Ernestopheles:
\Isotope\CompatibilityHelper::getProjectDir() ... coming later

system/modules/isotope/library/Isotope/Helper/Scope.php
Ernestopheles: Thanks, done

system/modules/isotope/library/Isotope/Helper/Scope.php
Ernestopheles: Thanks, done

system/modules/isotope/library/Isotope/Model/Attribute/AbstractAttributeWithOptions.php
```diff
@@ -117,15 +118,15 @@ public function getOptionsForWidget(IsotopeProduct $objProduct = null)
                     $arrOptions = array();
 
                 } elseif ($this->isCustomerDefined()) {
-                    $arrOptions = $objOptions->getArrayForFrontendWidget($objProduct, 'FE' === TL_MODE);
+                    $arrOptions = $objOptions->getArrayForFrontendWidget($objProduct,CompatibilityHelper::isFrontend());
```
aschempp:
```suggestion
                    $arrOptions = $objOptions->getArrayForFrontendWidget($objProduct, CompatibilityHelper::isFrontend());
```

system/modules/isotope/library/Isotope/Model/Attribute/AbstractAttributeWithOptions.php
```diff
@@ -138,7 +139,7 @@ public function getOptionsForWidget(IsotopeProduct $objProduct = null)
 
                 }
 
-                return $objOptions->getArrayForFrontendWidget($objProduct, 'FE' === TL_MODE);
+                return $objOptions->getArrayForFrontendWidget($objProduct,CompatibilityHelper::isFrontend());
```
aschempp:
```suggestion
                return $objOptions->getArrayForFrontendWidget($objProduct, CompatibilityHelper::isFrontend());
```

system/modules/isotope/library/Isotope/Model/Gallery/Standard.php
```diff
@@ -11,6 +11,7 @@
 
 namespace Isotope\Model\Gallery;
 
+use Isotope\CompatibilityHelper;
```
aschempp: Namespaces are usually sorted by name. My IDE (PHPStorm) does that automatically, did you manually add these?

system/modules/isotope/library/Isotope/Model/Gallery/Standard.php
```diff
@@ -66,7 +67,7 @@ class Standard extends Gallery implements IsotopeGallery
      */
     public function setRow(array $arrData)
     {
-        if ($arrData['customTpl'] != '' && 'FE' === TL_MODE) {
+        if ($arrData['customTpl'] != '' &&CompatibilityHelper::isFrontend()) {
```
aschempp:
```suggestion
        if ($arrData['customTpl'] != '' && CompatibilityHelper::isFrontend()) {
```

system/modules/isotope/library/Isotope/Module/WishlistDetails.php
```diff
@@ -73,7 +73,7 @@ protected function getCollection()
         }
 
         // Wishlist belongs to a member but not logged in
-        if ('FE' === TL_MODE && !$this->public && FrontendUser::getInstance()->id != $wishlist->member) {
+        if (CompatibilityHelper::isFrontend() && !$this->public && FrontendUser::getInstance()->id != $wishlist->member) {
```
aschempp, line 76: Looks like the `use` statement is missing in this class?

system/modules/isotope/library/Isotope/CompatibilityHelper.php
```diff
@@ -0,0 +1,78 @@
...
+ *
+ */
+class CompatibilityHelper
+{
+    /**
+     * Get the current request.
+     *
+     * @return Request|null
+     */
+    public static function getCurrentRequest()
```
aschempp: if the method is not used publicly, it should be private only for this class. We try not to expose unecessary API because people might use it and we'll never be able to change it afterwards.

aschempp (changes requested):

system/modules/isotope/library/Isotope/Model/Gallery/Standard.php
Ernestopheles: My IDE / ecs does that, too. But I deactivated formatting and did not run ecs because otherwise there would be several unwanted changes like removing spaces and other. As discussed before formatting shall therefore not be done by me. 
I am not happy with that, too, maybe there is a chance to customize my IDE and ecs?

system/modules/isotope/library/Isotope/CompatibilityHelper.php
Ernestopheles: Ok, that's a good reason to keep the method private. If it shows up being useful later, we can change it to public.

aschempp: exactly

system/modules/isotope/library/Isotope/Model/Gallery/Standard.php
aschempp: not sure there is such an option. If you manually added the imports, try adding them in alphabetical order. In PHPStorm there's an option to "Optimize Imports" which removes unused (as in your previous PR) and also orders them automatically.

Ernestopheles:
Such conflicts seems to be inevitable.
I usually create each PR based on branch 2.9. If one oder more PR are merged, this base is not valid any more for the remaining PR and might lead to conflicts as seen here.
Instead of creating independent PR maybe its better to base a PR on the last one created before? But this means, the PR must be merged in this sequence. I do not think we really do want that.
Not sure how to solve this. 
Hope its ok for you to resove this conflict (both use statements are wanted).

Ernestopheles:
What a mess! I did my best to resolve the conflicts; donna know why Track.php got all these unneccessary changes now; used contao\easy-coding-standard the way Contao Core does and only activated namespace ordering; this worked fine in all files except Track.php, where there are more (unneccessary) changes. Also did namespace ordering in more files (I hope this is ok).

Ernestopheles:
Please resolve this conflict

aschempp:
Thank you @Ernestopheles

PR #2500, opened 2023-11-06
Fix Unsupported operand types exception

koertho:
This PR fix the "Unsupported operand types" exception that appears if an product is added to cart (or cart is viewed) when the product has not price attribute. 

This PR adds just a float cast to the price calculation methods in `ProductCollectionItem`, so the price calculation methods always return a price :)

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/ProductCollectionItem.php
+++ b/system/modules/isotope/library/Isotope/Model/ProductCollectionItem.php
@@ -339,7 +339,7 @@ class ProductCollectionItem extends Model
      */
     public function getTotalPrice()
     {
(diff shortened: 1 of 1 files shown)


aschempp:
as discussed, it should return an empty string if there is no price on the product.

koertho:
Returning an empty string would only be possible for `getTotalOriginalPrice()`, the other methods are used follow-up price calculation methods. Switching to empty string lead to following exception:

image

koertho:
Therefore, since these three methods are price calculation methods, they also should IMHO return a price, even if the product itself has no price.

aschempp:
I have applied another fix in fda57f603b5eabcf989f07cf334d8c9269fd9387

PR #2501, opened 2023-11-09
Replace Haste\Units (Adaption to Contao5 Step 008)

Ernestopheles:
Replaced by jordanbrauer\unit-converter.

Fixed property type in 3 files.

Why draft?

1. Draft as PHPUnit Tests not yet written. Tested with Cypress with some simple usecases using the Isotope Demo. Includes only few usecases! More tests neccessary!!

2. Draft also as unsure if this solution fits in the projects concepts.

3. Draft also as being unhappy with not working DI (needed to use container->get).

4. Draft also as used PHP 8 typografy in constructors; therefore composer.json should reflect that (not yet included!)


aschempp:
I think the code pattern you should look at to fix the dependency injection issue is a "factory". Not sure if we need a `ScaleFactory` that creates new instanced of scales and a `WeightFactory` or maybe something combined. I would also not create services for all classes of the UnitConverter, I don't see any problem in hardcoding that in a factory method.

Ernestopheles:
Worked your advices in. Factory working, thanks for that advice! DI not working again. But solution working as should nethertheless.

If you want this solution, let me know and I will add PHPUnit tests for the new classes.

Allso let me know if the mass units that had been in Haste but are not in jordanbrauer should be added.

Remember I used PHP 8 notation in constructor.

Ernestopheles:
Is there anything left the project is waiting for?

aschempp:
mostly time on my side 😉

Ernestopheles:
So I am left here without answer. I suppose my work is not valuable for you?

aschempp:
This is Open Source software, you have no right to expect anything from anyone. Noone said your work is not valuable, but my private time is equally valuable to me. https://www.terminal42.ch/de/open-source

Ernestopheles:
Stopped working on adaption Isotope to Contao5 as I do not see any basis for cooperation with the project any more. What a pity!

Issue #2509, opened 2024-01-05
Add static type checking with phpstan


f-schnabel:
Hi I wanted to know if this repo also accepts Community Contributions.
If yes, I would like to contribute and add [phpstan](https://phpstan.org) as a static type checker into the project.
I would first start small with level 0 and gradually increase the strictness.

aschempp:
I guess that question has been answered 😅

PR #2510, opened 2024-01-06
Added phpstan level 0 and fixed various import issues

f-schnabel:
Part of #2509.
Adds phpstan with level 0, but there were some issues:
- Phpstan needs the swissbilling-api dependency to be installed, so I added it as a dev dependency
- In swissbilling-api there exists one class which is not psr-4 conform (ApiFactory) and this library also does not support php 8. I created a PR which also fixes these things (https://github.com/terminal42/swissbilling-api/pull/1).
- There was some issue with the FrontendTemplateTrait, which declared the getTemplate method as not static. I also created a PR for this (https://github.com/contao/contao/pull/6692)
- Phpstan often reportet "Unsafe usage of new static()." but I think this is intended, so I ignored this error for now. See also: https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /composer.lock
 /package-lock.json
 /.php_cs.cache
+/.idea
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@
 /composer.lock
 /package-lock.json
 /.php_cs.cache
-/.idea
--- a/composer.json
+++ b/composer.json
@@ -59,6 +59,7 @@
         "mpay24/mpay24-php": "^4.0",
         "friendsofphp/php-cs-fixer": "^2.14",
         "phpstan/phpstan": "^1.10",
+        "terminal42/swissbilling-api": "^1.0"
     },
     "suggest": {
         "terminal42/contao-fineuploader": "For file uploads in frontend using FineUploader",
--- a/composer.json
+++ b/composer.json
@@ -58,8 +58,8 @@
         "petschko/dhl-php-sdk": "dev-master@dev",
         "mpay24/mpay24-php": "^4.0",
         "friendsofphp/php-cs-fixer": "^2.14",
-        "phpstan/phpstan": "^1.10",
-        "terminal42/swissbilling-api": "^1.0"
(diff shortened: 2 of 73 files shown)


aschempp:
Thank you @Shadow-Devil !

f-schnabel:
I am a bit confused, since when I now run:
`vendor/terminal42/contao-build-tools/tools/phpstan/vendor/bin/phpstan analyze system`

There is no `#Function file_exists is unsafe to use#` matched, which you explicitly ignored. Can you please tell me why you added it?

aschempp:
There are two `file_exists` calls in `system/modules/isotope/library/Isotope/Template.php`. If I remove the `ignoreError`, I get that reported.

```
------ ------------------------------------------------------------------ 
  Line   isotope/library/Isotope/Template.php                              
 ------ ------------------------------------------------------------------ 
  50     Function file_exists is unsafe to use, rely on Symfony component  
         Filesystem::exists instead.                                       
  54     Function file_exists is unsafe to use, rely on Symfony component  
         Filesystem::exists instead.                                       
 ------ ------------------------------------------------------------------ 
```

aschempp:
btw, you can simply run `composer run phpstan` 😎

fritzmg:
`Filesystem::exists` is not safer than `file_exists`. It only introduces a check for `\PHP_MAXPATHLEN` - which `file_exists` does itself anyway. It looks like Symfony [introduced the check](https://github.com/symfony/symfony/commit/0d5f7e29050cf161dd620b4fa6d004d2cc7a81dc) in Symfony **2.3.38** (and later [updated it](https://github.com/symfony/symfony/commit/7a475595e2dd64790b3f1caf56f5b57cefcef27f) in Symfony **2.7.35** to use `\PHP_MAXPATHLEN`) due to [this PHP bug](https://bugs.php.net/bug.php?id=71103), which has since been fixed in PHP 7.1. The only advantage is that it will throw a nice `IOException` for you to catch in such a case.

aschempp:
Idk, I get the error when I run phpstan, I never asked for why it is there 😉

PR #2513, opened 2024-01-23
Remove type from model constructor argument

fritzmg:
Contao's base `\Contao\Model` class has no type on its constructor argument, as it allows both `\Contao\Database\Result` and `array`:

https://github.com/contao/contao/blob/7477d95ccb3cff5f5e315d9163904ebae8f1a0b3/core-bundle/src/Resources/contao/library/Contao/Model.php#L108-L110

This way you can pass an associative array from a database result (e.g. from `doctrine/dbal`) when instantiating a model. However, this is currently not possible with all of the Isotope models as they define the `\Contao\Database\Result` for the constructor argument. This PR fixes that.

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/Address.php
+++ b/system/modules/isotope/library/Isotope/Model/Address.php
@@ -66,7 +66,7 @@ class Address extends Model
      *
      * @param Result $objResult
      */
-    public function __construct(Result $objResult = null)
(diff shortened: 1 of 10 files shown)


fritzmg:
Rebased to `2.9` now.

aschempp (approved):

aschempp:
Thank you @fritzmg

PR #2514, opened 2024-01-29
Fix missing `tstamp`s for order stati

fritzmg:
When installing `isotope/isotope-core` for the first time, the order status back end view will look like this:

&nbsp;

&nbsp;
This is because `tstamp` is `0` for these database records. This PR fixes that.


fritzmg:
This is already fixed in `2.9`.

PR #2515, opened 2024-02-09
Add `doctrine/dbal` dependency

fritzmg:
Isotope uses Doctrine, but has no dependency currently. The dependency needs to be `^3.0` due to the usage of

```
Doctrine\DBAL\Connection::createSchemaManager()
```

for example.


aschempp:
Thank you @fritzmg

PR #2516, opened 2024-02-12
Fix type error in `BackendOverview`

fritzmg:
When using "Save & Duplicate" in a gallery config a type error will occur in `BackendOverview`.


aschempp:
Thank you @fritzmg

PR #2521, opened 2024-03-14
Feature/report sales total with conversation rate

error08:
Switch Sales Report to Apexcharts.
Add the feature conversation rate: If the visitors extension/bundle (https://github.com/BugBuster1701/contao-visitors-bundle) is installed, a visitor onfiguration can be created on the basis of which the conversation rate is calculated. If the extension is not installed, the report is displayed without the visitors and conversation rate. The extension is loosely coupled and therefore there are no direct dependencies to this bundle.

Bildschirm­foto 2024-07-26 um 13 18 14

Bildschirm­foto 2024-07-26 um 13 16 54


error08:
@aschempp Perhaps for version 2.9, too? Then the xcharts lib is replaced with apexcharts.

aschempp:
Thanks for working on this! I reverted the Javascript library, as it might be used by other people adding their own reports.

Checking the PR, there seems to be some custom implementation for a `tl_visitors` table? Not sure how that is related here?

error08:
@aschempp Sorry for my short description. I added some more details. I hope you will like that idea. When you have checked that add marked for 2.9, I will update the documentation for the reports.

aschempp:
I generally like the idea, but I think this should be a separate report (that is only available if you install the visitors bundle). It should not be part of the "default" in Isotope, that kinda makes the visitor bundle a default for Isotope, which I don't think it is.

error08:
This is how the implementation works. If the bundle is not installed, it looks as before, except that the combo box in the filter panel is always displayed. This could be changed so that it is only displayed when the bundle is installed. I find another report with almost identical data (if Visitor Bundle is installed) rather confusing.
That means if the bundle is not installed, it looks as before. No need to install anything.
I tried to implement this in the same way as with the DHL Business shipping method (https://docs.isotopeecommerce.org/manual/de/backend-configuration/store-configuration/checkout-flow/shipping-methods/shipping-method-dhl-business/), which is only displayed if a corresponding third-party extension is installed.

What might also make sense is to separate the two reports from the implementation. So that the report with the conversation rate functionality inherits from the previous report. And then the decision is made in the config.php as to which report is called behind the link & icon, depending on whether the Vistor Bundle is installed. This would make the implementation clearer.

aschempp:
It is similar since it shows similar data. But isn't that the case for all reports, since the show order data?

To me, this would be a report that reports on visitor configurations. Similar to reports that report on members etc.
DHL is a good example thought: the shipping method is a separate option, and it is only available if the dependencies are provided. Same should be here: the report should only be available if the module is installed. As a separate option.

error08:
If the report only shows visitors and the conversation rate, the connection to sales is unfortunately lost. I think this connection is particularly important here. The visitors column is not as important to display as the conversation rate column, but it also serves to improve understanding.

error08:
I will rework that and create a new pull request.

PR #2341, opened 2022-07-24
Move label generating logic in a separate method in `DC_ProductData`

ameotoko:
Fixes #2340

PR #2527, opened 2024-04-19
Add optional setting to prevent force-show first order column in backend

ameotoko:
This ports https://github.com/contao/contao/pull/6593 to Isotope's data container driver.

### DIFF
--- a/system/modules/isotope/drivers/DC_ProductData.php
+++ b/system/modules/isotope/drivers/DC_ProductData.php
@@ -1484,7 +1484,7 @@ class DC_ProductData extends \DC_Table
 <table class="tl_listing' . (($GLOBALS['TL_DCA'][$this->strTable]['list']['label']['showColumns'] ?? null) ? ' showColumns' : '') . ($this->strPickerFieldType ? ' picker unselectable' : '') . '">';
 
             // Automatically add the "order by" field as last column if we do not have group headers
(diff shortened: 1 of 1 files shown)


fritzmg (approved):

aschempp:
Can you please rebase this to the 2.9 branch? There are no plans to add more fixes to 2.8, and I think I have updated the driver for Contao 4.13 already (not sure if that is in there though).

ameotoko:
When will `2.9` be released then? Isotope always lacks compatibility with `4.13`. I already accepted that I don't have to learn Contao 5 for a long time, but support for `4.13` should be ongoing IMO.

fritzmg:
Both Isotope 2.8 and 2.9 are compatible with Contao 4.13.

ameotoko:
By "compatibility" I meant, that Isotope should not discard core's features when extending from Contao's core driver. This is currently the case in 2.8 with this issue and #2341 for example.

fritzmg:
Yes, but there won't be another 2.8 release, hence you should rebase this against 2.9.

ameotoko:
I'm a little confused now... So, if I found another bug (which I just did), I should submit the fix against `2.9`? But it means, I should be using `2.9` in production then? But it hasn't been released, it feels weird to use a dev branch in production, hence my question... Can you help me understand the process?

aschempp:
Thank you @ameotoko. I merged this into 2.9 now

PR #2534, opened 2024-08-23
Use path returned from hook for downloads

fritzmg:
The `downloadFromProductCollection` Isotope hook allows the path to be set as the return value - however, the path is then never actually used. This PR fixes that.


aschempp (approved):

aschempp:
Thank you @fritzmg

Issue #2536, opened 2024-09-07
Evaluations & statistics


katgirl:
Isotope Version 2.9.0

For sales statistics
![image](https://github.com/user-attachments/assets/a1b980a3-94cb-415a-a657-d8be67018436)

but at Sales summary
![image](https://github.com/user-attachments/assets/cbd5420a-97b9-4b65-b134-1715140a5dca)

The number of products is the number of units.
The sum of sales and the resulting average is also incorrect

katgirl:
The Problem is:

https://github.com/isotope/core/blob/2.9/system/modules/isotope_reports/library/Isotope/BackendModule/Reports.php#L150C24-L165C44

The ```LEFT JOIN``` at ```tl_iso_product_collection_item``` generate duplicate content by order.

aschempp:
@error08 could you help with this and tell if https://github.com/isotope/core/pull/2537 is the correct fix for your new statistics?

error08:
Hi @aschempp and @katgirl, if I understand it right, I would it solve this way: #2542. @katgirl: Your fix, sums the quantity of the items, but here we want display the products, right?

error08:
@katgirl The sales total here are the tax free value of the all paid orders in the time range minus the discounts. I will double check that next week.

aschempp:
> but here we want display the products, right?

If I understand correctly, we're not sure we want to show the number of "different products", or the number of quantities? Personally (I'm not a shop owner!) I would expect the number of items here, the number of things I will pack into my shipping boxes.

@katgirl did your statement about incorrect resulting average include net or gross prices?

aschempp:
Maybe we should name the columns "Net Sales Total" and "Net Sales ø" if that would make it more obvious?

aschempp:
Any opinions on the matter?

error08:
I'm fine with "Net Sales Total" and "Net Sales ø". Regarding the issue, I will try to test it this weekend.

error08:
@aschempp , @katgirl : I have found the bug, please double check the PR [2542](https://github.com/isotope/core/pull/2542)

aschempp:
see https://github.com/isotope/core/pull/2542

PR #2537, opened 2024-09-07
Update Reports.php

katgirl:
see #2536


aschempp:
see https://github.com/isotope/core/pull/2542

PR #2541, opened 2024-10-08
Use URL from defined root in category filter

fritzmg:
If you defined a root page in the category filter, the menu will be built according to that root, but the `href` of each item will reference the current page instead of the defined root.

I am not sure this is intended? If not - this PR would fix that by always generating the URL for the defined root instead.

### DIFF
--- a/system/modules/isotope/library/Isotope/Module/CategoryFilter.php
+++ b/system/modules/isotope/library/Isotope/Module/CategoryFilter.php
@@ -208,7 +208,7 @@ class CategoryFilter extends AbstractProductFilter implements IsotopeFilterModul
                 'categoryfilter=' . $value,
                 Url::removeQueryStringCallback(function ($value, $key) {
                     return strpos($key, 'page_iso') !== 0;
-                })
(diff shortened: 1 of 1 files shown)


aschempp:
Hmm, I do think that is intended. If you have a filter for categories = pages on the home page, you don't want to get redirect to that page if you apply the filter. You want the product list on that (home) page to be filtered. If you want a redirect, you can simply use a navigation module for the same purpose, right?

fritzmg:
> Hmm, I do think that is intended. If you have a filter for categories = pages on the home page, you don't want to get redirect to that page if you apply the filter. You want the product list on that (home) page to be filtered.

True, yeah, then this should not be merged.

> If you want a redirect, you can simply use a navigation module for the same purpose, right?

No, the navigation module would not filter anything.

aschempp:
> No, the navigation module would not filter anything.

Well if you assign products to categories (pages) and have list modules on them, each page would only show these products, similar to the category filter.

fritzmg:
I suppose so, yes. In our case we have a mix of both.

PR #2545, opened 2024-10-28
Take hash URLs into consideration and fix double ampersand

fritzmg:
There are two issues with the current `mod_iso_productlist_caching` template:

#### Hash URLs

If the current URL contains a hash, the current logic will create an URL like this for the AJAX request:

```
https://example.com/products?foobar=1#foobar&&buildCache=1
```

This will result in the `buildCache` parameter being ignored and thus it will only fetch a state where "Loading products ..." is displayed again.

#### Double Ampersand

The current

```js
(document.location.search ? '&' : '?') + '&buildCache=1'
```

results in URLs like 

```
https://example.com/products?foobar=1&&buildCache=1
```

This PR fixes both.

### DIFF
--- a/system/modules/isotope/templates/modules/mod_iso_productlist_caching.html5
+++ b/system/modules/isotope/templates/modules/mod_iso_productlist_caching.html5
@@ -12,9 +12,11 @@
     (function() {
         Isotope.displayBox('<?= $this->message ?>');
 
-        var url = window.location.href + (document.location.search ? '&' : '?') + '&buildCache=1',
-            xhr = new XMLHttpRequest();
+        var url = window.location.href;
+        url = url.substr(0, url.indexOf(window.location.hash) || url.length);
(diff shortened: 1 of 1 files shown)


fritzmg:
Note: there are a bunch of different ways to get or generate the current URL without the `#` hash part - feel free to adjust to your preferred method 😁

aschempp:
I think it makes more sense to use the classes available, see f3ac4d5f6e75cf7aabb0a358dfca92291e39a7a6 for my fix. Hope that works for your case as well 😅

fritzmg:
> I think it makes more sense to use the classes available, see [f3ac4d5](https://github.com/isotope/core/commit/f3ac4d5f6e75cf7aabb0a358dfca92291e39a7a6) for my fix. Hope that works for your case as well 😅

That should work in principle, but there is an error in your commit. See https://github.com/isotope/core/pull/2547

PR #2547, opened 2024-10-31
Fix missing bracket in `mod_iso_productlist_caching`

fritzmg:
This syntax error was introduced in https://github.com/isotope/core/commit/f3ac4d5f6e75cf7aabb0a358dfca92291e39a7a6


aschempp:
Thank you @fritzmg

PR #2549, opened 2024-11-04
Remove `patchwork/utf-8` dependency since it is abandoned

### LINKED ISSUE #2548: Replace abandoned patchwork/utf8 dependency
It would be great to replace the [patchwork/utf8](https://packagist.org/packages/patchwork/utf8) dependency since it is abandoned.
It is only used four times in the [DC_ProductData](https://github.com/isotope/core/blob/46fb9d424e063ad914c1131fcb369a3478cd41bb/system/modules/isotope/drivers/DC_ProductData.php) file to sort some `$options_sorter`:
```php
        // Sort by option values
        uksort($options_sorter, array(Utf8::class, 'strnatcasecmp'));
```

There also exists the [strnatcasecmp](https://www.php.net/manual/en/function.strnatcasecmp.php) function directly in PHP, but I am unsure if it works the same for non ASCII characters (and if thats even relevant).

If this can simply be replaced then I can also create a PR for this.

f-schnabel:
Fixes #2548

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,6 @@
         "terminal42/notification_center": "^1.0",
         "terminal42/dc_multilingual": "^4.0",
         "tecnickcom/tcpdf": "^6.2.22",
-        "patchwork/utf8": "^1.2",
         "symfony/filesystem": "^5.4 || ^6.0",
         "symfony/http-client": "^4.4 || ^5.0 || ^6.0",
         "symfony/polyfill-php80": "^1.25",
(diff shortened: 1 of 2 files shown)


fritzmg (changes requested):
The `symfony/string` dependency needs to be added as well.

f-schnabel:
@fritzmg thank you. Just `5.4^` or `5.4^ || 6.0^`?

fritzmg:
I suppose just `^5.4` will suffice, since Isotope 2.x will not receive a release for Contao 5+. But technically you can allow higher versions, if compatible.

fritzmg (approved):

aschempp:
Thanks for your pull request! I'm sorry I didn't see it early enought and fixed this myself in 31bcc536e22980148bc71de4744653f350f8c881 🙈

PR #2550, opened 2024-11-18
Remove `tableless`

fritzmg:
The `tableless` setting was removed in Contao 4.0.0. Isotope still references it - but since Isotope now requires at least Contao 4.13, it can be removed.


fritzmg:
@aschempp this can be safely merged into `2.9`. Contao 4 has no support for `tableless`, so all the `tableless` options within Isotope are superfluous.

aschempp:
I know, but I don't see any point in changing 2.9 for something that has no effect. Its easy to just remove it in the next major version.

PR #2552, opened 2024-11-29
Fix product Standard::getVariantIds()

### LINKED ISSUE #2551: Product::getVariantIds() not working for variants with guests property
Isotope: 2.9.4

When using a product type with variants, having the property `guests` activated for variants, the function `Isotope\Model\ProductStandard::getVariantIds()` does not work correctly.
Variant ids are loaded by an sql query, but the statement misses the property geusts:

```php
$strQuery = 'SELECT tl_iso_product.id, tl_iso_product.protected, tl_iso_product.groups FROM ...';
...
$objVariants = Database::getInstance()->query($strQuery);
```

The property guests must be added to the statement because it is used in the variant validation, resulting in the variant not being validated:

```php
if (... && (!$blnHasGuests || !$objVariants->guests)) {
    ...
```

markom-werbeagentur:
- Fix https://github.com/isotope/core/issues/2551 (line 264)
- Fix product variants with guests=1 having protected property not being visible for logged in members having a matching group (line 303)
- Fix runtime error when a protected product variant having guests=1 for guest users (line 308)

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/Product/Standard.php
+++ b/system/modules/isotope/library/Isotope/Model/Product/Standard.php
@@ -261,7 +261,7 @@ class Standard extends AbstractProduct implements WeightAggregate, IsotopeProduc
             $blnHasGuests    = false;
             $isMember = \Contao\System::getContainer()->get('security.helper')->isGranted('ROLE_MEMBER');
             $strQuery        = '
-                SELECT tl_iso_product.id, tl_iso_product.protected, tl_iso_product.groups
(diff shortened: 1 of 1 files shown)


aschempp (approved):

aschempp:
Thank you @markom-werbeagentur!
I re-added your change in 89a483258ef5ec4169817eb16a71ae69e08970b1 😏

markom-werbeagentur:
Thanks for the implementation :)

PR #2553, opened 2024-12-02
Add HTTP cache tags

fritzmg:
Currently, if a page with a product list is in the shared HTTP cache and a new product is added, deleted or a product is changed, the cache will not be invalidated. Likewise if a product detail is in the shared HTTP cache and that product is changed, the shared cache is also not invalidated.

This PR fixes that by adding at least the general `contao.db.tl_iso_product` cache tag for any product list and the `contao.db.tl_iso_product.*` cache tag for any product detail. Both will get automatically invalidated by the `DataContainer` in the back end when editing products.

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/ProductCache.php
+++ b/system/modules/isotope/library/Isotope/Model/ProductCache.php
@@ -15,6 +15,7 @@ use Contao\Database;
 use Contao\FrontendUser;
 use Contao\Input;
 use Contao\Model;
+use Contao\System;
 
 /**
  * Isotope\Model\ProductCache represents an Isotope product cache model
@@ -227,6 +228,10 @@ class ProductCache extends Model
     public static function purge()
     {
         Database::getInstance()->query("TRUNCATE " . static::$strTable);
+
+        if (System::getContainer()->has('fos_http_cache.cache_manager')) {
+            System::getContainer()->get('fos_http_cache.cache_manager')->invalidateTags(['contao.db.tl_iso_product']);
+        }
     }
 
     /**
--- a/system/modules/isotope/library/Isotope/Module/ProductList.php
+++ b/system/modules/isotope/library/Isotope/Module/ProductList.php
@@ -303,6 +303,12 @@ class ProductList extends Module
         ;
 
         $this->Template->products = $arrBuffer;
+
+        // Add cache tag
+        if (System::getContainer()->has('fos_http_cache.http.symfony_response_tagger')) {
+            $responseTagger = System::getContainer()->get('fos_http_cache.http.symfony_response_tagger');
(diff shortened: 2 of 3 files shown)


aschempp:
This might partially work, but I would suggest to at least also also add cache clearing to `Backend::truncateProductCache`

fritzmg:
> This might partially work

Partially is still better than never 😁.

> but I would suggest to at least also also add cache clearing to Backend::truncateProductCache

Makes sense 👍. I will add it to the PR.

fritzmg:
I have added it to `ProductCache::purge()` instead, wdyt?

system/modules/isotope/library/Isotope/Model/ProductCache.php
```diff
@@ -227,6 +228,10 @@ public static function deleteForPageAndModuleOrExpired($intPage, $intModule)
     public static function purge()
     {
         Database::getInstance()->query("TRUNCATE " . static::$strTable);
+
+        if (System::getContainer()->has('fos_http_cache.cache_manager')) {
+            System::getContainer()->get('fos_http_cache.cache_manager')->invalidateTags(['contao.db.tl_iso_product']);
```
aschempp, line 233: this would not clear the reader pages though, right?

fritzmg: Only the product list uses the `ProductCache`.

aschempp: yes, but this method is called all over the place where the cache of products should be cleared. E.g. if the store configuration is changed, the reader pages might display a wrong currency. So it would kinda make sense to clear them as well? Although we would be better off actually implementing the correct tags. Which is why this might not make sense as a quick fix in 2.9 🙈

fritzmg: > Although we would be better off actually implementing the correct tags.

I can add the additional cache tags in this PR if you want.

> Which is why this might not make sense as a quick fix in 2.9

Why? It's a bug in 2.9 and can easily be fixed there.

aschempp: idk, its also a feature of Contao that is not supported by Isotope 🤷

fritzmg: HTTP caching you mean?

aschempp: cache tags

fritzmg: But again: it's better than to clear the cache not at all, is it not? Having this issue again in a project.

Issue #2559, opened 2025-01-08
negative discount for rule leads to floating point error


xprojects-de:
Isotope 2.9.6, C4.13, PHP 8.1 and also PHP 8.3

A discount with a negative percentage value will randomly lead to incorrect calculations (see screenshots. 139: wrong, 159 ok)

![Bildschirmfoto 2025-01-08 um 14 25 41](https://github.com/user-attachments/assets/82172c19-ee71-40bf-bc1c-ba294e56ea00)
![Bildschirmfoto 2025-01-08 um 14 26 01](https://github.com/user-attachments/assets/596dc329-9623-44cf-a221-0a92a5458ecb)

The case here is a product price of 139.00 and a discount of -10% as rule for each Product and normal rounding.
![Screenshot 2025-01-08 at 14-44-22 Regeln   Gutscheine Bearbeite die Regel mit der ID 9 bergwerk net](https://github.com/user-attachments/assets/8e67df2e-8d2c-427c-886a-83b75f6fd9d4)

After testing and discussion it seems to be a floating point error for ceil at https://github.com/isotope/core/blob/14c204b59d756cf3400a8d19f804da1d7dd99348/system/modules/isotope_rules/library/Isotope/Model/ProductCollectionSurcharge/Rule.php#L205

As @fritzmg said the error can be reproduced using
```php
$fltPrice = (139.0 / 100) * -10.0;

var_dump([
    $fltPrice,
    $fltPrice * 100,
    ceil($fltPrice * 100),
    ceil($fltPrice * 100) / 100,
]);
```

More test results:

dd at line 205:
```php
dd(
    $fltPrice,
    $fltPrice * 100,
    ceil($fltPrice * 100),
    ceil($fltPrice * 100) / 100,
);

// Result:
^ -13.9
^ -1390.0
^ -1389.0
^ -13.89
^ -13.9
^ -1390.0
^ -1389.0
^ -13.89
```

ausi:
You can round it to the desired precision beforehand (maybe only directly before calling floor or ceil). But it depends on what the code actually wants to have as a result in certain edge cases...

```php
$fltPrice = (139.0 / 100) * -10.0;
$fltPrice = round($fltPrice, 5);

var_dump([
    $fltPrice,
    $fltPrice * 100,
    ceil($fltPrice * 100),
    ceil($fltPrice * 100) / 100,
]);

// array(4) {
//   [0]=>
//   float(-13.9)
//   [1]=>
//   float(-1390)
[...]

PR #2562, opened 2025-01-20
fix floating point error for rule discount

xprojects-de:
see https://github.com/isotope/core/issues/2559#issuecomment-2580593127

I chose a round precision of 4 because this has already been used at https://github.com/isotope/core/blob/d72bb80f13d014f609a0dea8cd8c3bb84fe6c5da/system/modules/isotope_rules/library/Isotope/Model/ProductCollectionSurcharge/Rule.php#L237

Feel free to modify :-)

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/ProductCollectionSurcharge.php
+++ b/system/modules/isotope/library/Isotope/Model/ProductCollectionSurcharge.php
@@ -165,7 +165,7 @@ abstract class ProductCollectionSurcharge extends TypeAgent
                 $fltProductPrice = $this->total_price / 100 * (100 / $fltTotal * $objItem->getTaxFreeTotalPrice());
             }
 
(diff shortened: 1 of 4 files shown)


xprojects-de:
Having a price of 139,00 with a discount of -10% the result is 13,90.
Having a price of 139,90 with a discount of -10% the result is always 13,99!
Having a price of 139,91-139,94 with a discount of -10% the result is 13,99 for "normal" and "down" and 14,00 for "up"!
Having a price of 139,95-139,99 with a discount of -10% the result is 13,99 for "down" and 14,00 for "normal" and "up"!

Do you think this is ok or what would be the correct value here?

aschempp:
Sorry for taking this long. I have finally taken the time to look into this, and refactored the calculation function to the rule model, so the code only exists once. My manual tests showed this still works as expected, but can you please check as well?

Also be aware that the issue does not seem to exist if you would use the rule on cart total option 🙈

xprojects-de:
seems to work. Thanks!

aschempp:
Thank you @xprojects-de

PR #2565, opened 2025-02-14
Do not hard code `.tl_filter` width

fritzmg:
If you add `'sorting' => true` to more fields of `tl_iso_product_collection` in a DCA adjustment, the filters will currently overflow:

image

This PR fixes that by setting the `width` of `.tl_filter` to `auto` instead:

image


aschempp:
Any idea what the reason was to add this in the first place?

fritzmg:
No, can't think of any reason to set the width this way.

aschempp (approved):

aschempp:
Thank you @fritzmg

PR #2570, opened 2025-04-16
Fix possible runtime error when product is not registered

markom-werbeagentur:
`ProductCategory::findByPidForPublishedPages($productId)` causes a runtime error when the given product is not registered in the model registry:

```
Cannot instantiate abstract class Isotope\Model\Product
in vendor/contao/core-bundle/src/Resources/contao/library/Contao/Model.php (line 182)
```

`Contao\Model::__construct()` tries to load the relations for `tl_iso_product_category.pid`, so it instanciates `Isotope\Model\Product` skipping the type agent for `tl_iso_product`.

Solution: register the product.

This issue does not appear in vanilla isotope, but may appear when you have custom bundles extending isotope.
Reproduced in Contao 4.13.

Steps to reproduce:
```php
$product = Product::findByPk($myId);
$product->detach(false);
ProductCategory::findByPidForPublishedPages($myId);
```

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/ProductCategory.php
+++ b/system/modules/isotope/library/Isotope/Model/ProductCategory.php
@@ -63,8 +63,9 @@ class ProductCategory extends Model
      */
     public static function findByPidForPublishedPages($intProduct, array $arrOptions = array())
     {
-        // Register the model to prevent a runtime error
+        // Register model to prevent eager loading of product data (see isotope/core#2570)
         Product::findByPk($intProduct);
+
         $arrOptions = static::getFindByPidForPublishedPagesOptions($intProduct, $arrOptions);
 
         return parent::find($arrOptions);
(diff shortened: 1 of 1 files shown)


aschempp:
Do I understand correctly that you are manually instantiating a product model without a type? Why would you do that?

markom-werbeagentur:
No I did not do that. Please see `Steps to reproduce` - thats what I did, while `$myId` is a valid product id. The product must also have a published category page in order to reproduce the issue. Everything above is just what happens when you do so.

`ProductCategory::findByPidForPublishedPages($myId);` causes the runtime issue when there is no product with the id `$myId` in the model registry, beacause `ProductCategory::__construct()` automaticly tries to load relations on `tl_iso_product_category.pid`, inherited from `Contao\Model`. I hope this description is clearer.

> Do I understand correctly that you are manually instantiating a product model without a type? Why would you do that?

aschempp (approved):
I see. The problem is eager loading, which is necessary to check the page data though.

aschempp:
Thank you @markom-werbeagentur

PR #2574, opened 2025-06-18
Fix product collection model type

markom-werbeagentur:
### Problem

The TypeAgent breaks when you try to extend any subclass of ProductCollection. Example:

```php
// config.php
\Isotope\Model\ProductCollection::registerModelType('cart', \MyCustomNamepsace\Cart::class);
```

Doing so leads to an error when Isotope tries to create a new cart for the current user:
`Isotope\Model\ProductCollection\Cart (tl_iso_product_collection.) has no model type`. The actual issue is in `\Isotope\Model\ProductCollection\Cart` line 348:

```php
if ($objCart === null) {
    $objConfig = Config::findByRootPageOrFallback($objPage->rootId);
    $objCart   = new static();
    // .......
```

Calling `new static()` creates a new instance of type agent, that is not a valid model type anymore, because `\MyCustomNamepsace\Cart::class` should be the model type.

### Expected behavior

The ProductCollection should dynamicly load the correct model type and the related class.

```php
// Pseudo code
$objCart   = new {static::getClassname()} ();
```

### Suggestion

Create a new property `ProductCollection::$modelType` and set it correctly in each class. Example for `Cart`:

```php
protected static $modelType = 'cart';

// ....
public static function findForCurrentStore()
{
	// ...
	if ($objCart === null) {
		$objConfig = Config::findByRootPageOrFallback($objPage->rootId);
		$objCart   = new (static::$arrModelTypes[static::$modelType])();
```

### DIFF
--- a/system/modules/isotope/library/Isotope/Model/ProductCollection.php
+++ b/system/modules/isotope/library/Isotope/Model/ProductCollection.php
@@ -1955,7 +1955,7 @@ abstract class ProductCollection extends TypeAgent implements IsotopeProductColl
      */
     public static function createFromCollection(IsotopeProductCollection $objSource)
     {
-        $objCollection = new static();
+        $objCollection = new (static::$arrModelTypes[static::$modelType])();
         $objConfig = $objSource->getConfig();
 
         if (null === $objConfig) {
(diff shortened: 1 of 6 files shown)


aschempp:
`TypeAgent` are not meant to be replaced (especially not the core ones) in Isotope 2. This is a quite fragile setup, not everything has interfaces etc.
I would not consider this a bug but rather a feature. And Isotope 2 is feature-complete (there will not be a new minor version). So far it looks like this will be possible in Isotope for Contao 5 though 😎

markom-werbeagentur:
Thanks for the reply. I did a workaround anyway that does not require extending the cart.
Looking forward to Isotope 3 for Contao 5 :)

PR #2578, opened 2025-08-14
Update ProductCollectionSurcharge Rule.php

tsarma:
Source Id for a rule is always stored as 0, inside table `tl_iso_product_collection_surcharge`. With this change source_id is stored correctly.
screenshot_2021-02-16_at_11 48 56


aschempp (approved):

aschempp:
Thank you @tsarma

PR #2579, opened 2025-09-29
Add order state for free orders

error08:
Adding a field to set the status of free orders so that it can be done flexibly, decoupled from the status of new orders.


aschempp:
Unfortunately, we will not be adding new features to Isotope 2.9, therefore I can't merge this as-is. I'll keep it open to think about it in a future version though 😊

PR #2580, opened 2025-11-03
Fix some warnings

michb:
- `Undefined array key "format"` in attribute model while generating price tiers
- `Undefined array key "noHtml"` in file tree attribute model while generateAttribute


aschempp (approved):

aschempp:
Thank you @michb

PR #2581, opened 2025-12-01
Add a null check to prevent an undefined array key warning

bytehead:
My sentry gets flooded with `undefined array key ""` warnings.

Bildschirmfoto 2025-12-01 um 08 54 53


fritzmg (approved):

aschempp (approved):

aschempp:
Thank you @bytehead

Issue #2251, opened 2021-09-16
tax_free_price in tl_iso_product_collection_item when using Rule is wrong


xApep:
Hey.
I have opened similar issue about that before but it is not so much related to that.

After Order is done, I'm sending order data to some 3rd party software. 
I'm accessing order details with ProductCollection\Order and getItems().

_**issue: tax_free_price is not calculated correctly when product have rule.**_

Real case example:
Product gross price before discount: **38€**
Discount value: **5€**
Product gross price after discount: **33€**
Tax rate: 9.5%

In table tl_iso_product_collection_item, tax_free_price is set to 29.70
Isotope calculation: 
Tax amount: 38 - 9.5% (3.3) = 34.7
tax_free_price = 33 - 3.3 = 29.70

**but this is wrong** because it deduct tax amount from gross price but not tax amount of discounted price.

**So the correct value of tax_free_price would be:** 
product_price (gross price after discount, field price in collection_item) - tax rate of that price = tax_free_price
33 - 9.5% (2.86) = **30.14**

In the tl_iso_product_surcharge I can see that the total_price matches the 2.86, but it is not used in calculation of tax_free_price.

_Tested Isotope Version: 2.6.15 and 2.7.3_

aschempp:
Fixed in 464717611395c0cc403b7bf99aa33dc484aa2059 🎉

PR #2584, opened 2026-09-17
Fix regression introduced for the fix for #2251

jaapjansma:
The fix for #2251 works as expected. However a `+` is missing on https://github.com/isotope/core/blob/2.9/system/modules/isotope_rules/library/Isotope/Model/ProductCollectionSurcharge/Rule.php#L221

This fixes that. 

The problem is when you have a rule which applies to the sub total of your cart and the rule also has VAT. Then the tax_free_total_price is wrongly calculated. Only of the last item in the cart is used in the calculation.


aschempp (approved):

aschempp:
Thank you @jaapjansma

